PR #556, opened 2013-11-25
systempatches, allow patches with a different folder structure to be committed (useful for packages) v2

PiBa-NL:
systempatches, allow patches with a different folder structure to be committed (useful for packages) v2

Same purpose as for https://github.com/pfsense/pfsense-packages/pull/555 but now improved like this:
-no unnecessary whitespace changes
-'simplified' the solution for the parsing/modification of the patch paths.

The intention of the patch is that when you 'fetch' a change from github even if the paths are different its still possible to apply the patch. So when both config/systempatches/patches.inc and config/systempatches/system_patches.php  are affected it is possible to apply the patch. To /usr/local/www/system_patches.php and /usr/local/pkg/patches.inc without the need to manually split up the patch.

Like it was before, the 'modification routine' is optional and must be chosen with a checkbox for a patch to be affected. So their is no possibility for negative affects on current existing and working patches.

@jim-p, i hope this one is ok, as creating 'proper' patches and then using gist to distribute them seams troublesome to me when using git for sourcecontrol but needing to move files around before creating the patch file.. (or needing to manually modify the patches each time). I use the patches package a lot during development&testing and also to easily get the latest changes into my production machine with an easy way to revert any issues that might sometimes arise. I found that creating separate patches for each affected file, (as i cant select 2 out of 3 changed files to create a patch for), then saving them all in the patches package then applying them all is a rather error prone procedure..

Or perhaps should each package represent the directory tree or part thereof it uses? So i should for haproxy move some files to /config/haproxy-devel/www/ and /config/haproxy-devel/pkg/ instead of having them all in 1 directory?

### DIFF
--- a/config/systempatches/patches.inc
+++ b/config/systempatches/patches.inc
@@ -34,6 +34,7 @@ $git_root_url = "http://github.com/pfsense/pfsense/commit/";
 $patch_suffix = ".patch";
 $patch_dir = "/var/patches";
 $patch_cmd = "/usr/bin/patch";
+$patches_config = &$config['installedpackages']['patches'];
 
 function patch_package_install() {
 	patch_add_shellcmd();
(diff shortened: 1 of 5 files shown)


cbuechler:
This now has merge conflicts, and 2.2.x and earlier packages are no longer actively maintained

PR #631, opened 2014-03-21
Add option "fast hello"

pkilla:
ip ospf dead-interval minimal hello-multiplier <2-20>

dead-interval is set to 1 second
hello-multiplier specifies how many "hellos" to send per second

Signed-off-by: pkilla <pkilla@yandex.ru>


rbgarga:
Almost 10 months and CLA was not signed. Closing.

PR #641, opened 2014-04-07
haproxy-devel, support transparent-proxy with IPv6 and FreeBSD10 (will need a separate kernel patch for 'divert-reply' support.)

PiBa-NL:
haproxy-devel, support transparent-proxy with IPv6 and FreeBSD10 (will need a separate kernel patch for 'divert-reply' support.)

### DIFF
--- a/config/haproxy-devel/haproxy.inc
+++ b/config/haproxy-devel/haproxy.inc
@@ -609,8 +609,12 @@ function write_backend($fd, $name, $pool, $frontend) {
 		$pool['retries'] = 3;
 	fwrite ($fd, "\tretries\t\t\t" . $pool['retries'] . "\n");
 
-	if ($pool['transparent_clientip'])
-		fwrite ($fd, "\tsource 0.0.0.0 usesrc clientip\n");
+	if ($pool['transparent_clientip']) {
+		if (is_ipaddrv4($frontend_ip))
+			fwrite ($fd, "\tsource 0.0.0.0 usesrc clientip\n");
+		else
+			fwrite ($fd, "\tsource ipv6@ usesrc clientip\n");
+	}
 		
 	if($pool['stats_enabled']=='yes') {
 		fwrite ($fd, "\tstats\t\t\tenable\n");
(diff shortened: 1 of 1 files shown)


rbgarga:
Is this still relevant @PiBa-NL?

PiBa-NL:
Hi @rbgarga , whell.. the 'divert-reply' functionality would likely work better/faster than the current combination of stateless and ipfw rules.. Or it would at least be a 'cleaner' solution.. As pf is always loaded in pfsense and ipfw normally only by captive portal..
Anyway for this commit to be usefull first the divert-reply functionality must be added to pf.
I've added a redmine https://redmine.pfsense.org/issues/3943 issue and patch (under the freebsd link) for that a while back. I would be interested in your opinion about it, thanks in advance.

PiBa-NL:
b.t.w. I cant remember if i have checked if transparent-client-ip works with ipfw and ipv6..

cbuechler:
Has conflicts and 2.2.x and earlier packages are no longer actively maintained.

PiBa-NL:
I can resubmit it for 2.3, but it would be nice if some form of divert-reply could be implemented in pf. Any advice on how to make that happen.? The freebsd patch is submitted but if there is no chance it will be commited either upstream or on pfSense specifically then i dont need to spend time to rebase it for current version.. What do you advice.?

gonzopancho:
If there is a freebsd patch, and we can take it without impacting other
functionality, we'll get it into pfSense.

I suggest we work together to make this happen.

JIm

cbuechler:
The patch is on this PR. Not sure how good or complete it might be, but this would be nice functionality to have. 
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=188511

PiBa-NL:
The patch did work at that time..
p.s. for your information, Kristof (pf maintainer) was going to take a look at the patch but Ermal commented on mailinglist "the better implementation in general, which is on my roadmap, is convert the route-to/reply-to" but i have not seen if he made any progress with that. Nor think i would be able to make that proposed change.. https://lists.freebsd.org/pipermail/freebsd-pf/2015-August/007834.html

PR #666, opened 2014-06-04
Change way to stop openbgpd and add status on rc file bgpd.sh

SilvioGiunge:
The bgpd contains two linked process with him, In the rc file openbgpd is stoped by an killall command,  when the command kill the session engine, he automatically stop the another processes, but killall command try to stop the another processes too, generating an error on system.log file.

Trying to stop the session engine by the 'kill -15' command, and to verify if another processes is running before to use the killall command can be a better way to stop the bgpd.

Add status option in the rc file can help to see if the service is running.''

### DIFF
--- a/config/openbgpd/openbgpd.inc
+++ b/config/openbgpd/openbgpd.inc
@@ -189,36 +189,82 @@ function openbgpd_install_conf() {
 	unset($conffile);
 
 	// Create rc.d file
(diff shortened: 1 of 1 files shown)


rbgarga:
Please re-submit with a simpler rc.d script as mentioned on last comment

PR #676, opened 2014-06-17
Update haproxy.inc

ceylani:
getting IPs from alias table added


PiBa-NL:
@ceylani i think this pullrequest is nolonger needed? if you agree please close it.

rbgarga:
Looks like it's no longer needed. If you think it's still relevant, please re-submit.

PR #677, opened 2014-06-17
Update haproxy_utils.inc

ghost:
IP alias list added to external address combo


PiBa-NL:
@ceylani i think this pullrequest is nolonger needed? if you agree please close it.

PR #686, opened 2014-07-19
Handle port aliases & validate port on use

stilez:
Anyterm settings accept textual input for the port (useful: an alias can be specified, e.g., for fw ACL rules). 

But the code for the actual service errors out nastily if an alias or non-port is used, since it assumes $port exists, is valid, and is already numeric.  None of these are necessarily true, it should instead check the data is meaningful and resolve if a port alias, before using raw text.

Tested, works.  Effects of patch:

1) explicitly checks a port is defined;
2) correctly handle port if textual (i.e., port alias) and validate value obtained (numeric 1-65535) within both server and client code
3) provides at least rudimentary error handling for invalid data (like Unbound, an attempt to use the functional tab if data is bad, returns the user to the package "Settings" tab instead).

NOTE - there is a small alias handling issue revealed by this - if you define an alias "AnytermPort" as port 999, then the contents of $config['aliases']['AnytermPort'] is left padded with spaces ("  999" not "999"), and filter_expand_alias() returns this value which fails since (is_numericint($port) && $port <-=65535) isn't true due to padding. Workaround - using intval() first, since the fix is in alias handling not here.

### DIFF
--- a/config/anyterm/access_anyterm.php
+++ b/config/anyterm/access_anyterm.php
@@ -27,18 +27,31 @@
     POSSIBILITY OF SUCH DAMAGE.
 */
 
-require("guiconfig.inc");
+require_once("guiconfig.inc");
+require_once("filter.inc");  // needs filter_expand_alias() in case anyterm port is an alias
 
 if($config['installedpackages']['anyterm']['config'][0]['stunnelport']) {
 	$port = $config['installedpackages']['anyterm']['config'][0]['stunnelport'];
 	$httpors = "https";
-} else {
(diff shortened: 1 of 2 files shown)


rbgarga:
@stilez can you please rebase your fork? This cannot be automatically merged anymore.

stilez:
Looks like anyterm has been removed - I can't seem to find the files these patches should be applied to, at the moment. What do you want me to do?

rbgarga:
Right, sorry for the noise. Closing it!

PR #701, opened 2014-09-16
tinc, fixes re-installation, while performing a pfSense firmware upgrade

PiBa-NL:
tinc, fixes re-installation, while performing a pfSense firmware upgrade


jim-p:
Merged this earlier this week.

PR #711, opened 2014-09-28
spamd, fix spamtrap and blacklist buttons, fix startup on pfSense 2.2

PiBa-NL:
spamd, fix spamtrap and blacklist buttons, fix startup on pfSense 2.2


rbgarga:
It cannot be automatically merged, could you please rebase your fork?

PR #714, opened 2014-10-08
Add Recursive Search User in LDAP Auth

djgel:
Add Recursive Search User in LDAP Auth


rbgarga:
Squid3-dev was removed from packages on 2.2, this is not relevant anymore.

PR #747, opened 2014-12-08
zabbix2 - update packages to 2.4.2 - first step

dbaio:
Preparing for upgrade.
Need to build all binaries: zabbix2 agent and proxy


rbgarga:
Please do not change it on pkg_config.8 files, this build will not be possible for 2.1. Also, you can just remove <build_port_path> parameter since it is not necessary for PBIs

dbaio:
Done. Thanks @rbgarga

PR #776, opened 2015-01-09
HAVP configurator "Log" option now really using

SunStroke74:
HAVP webConfigurator "Log" checkbox not using at all. Error logs was written always, regardless of "Log" option. Change this behaviour to reflect configurator option. Note: HAVP access log writes always because of virus attentions.


rbgarga:
Please bump package version

PR #782, opened 2015-01-10
update squidguard-dev to stay consistent with squidguard

CinoNY:
updated squidguard-dev to stay consistent with changes that were made squidguard


CinoNY:
after speaking with jim, canceling this request

PR #783, opened 2015-01-10
OpenVPN export fix index typo

phil-davis:
The $server array here has and "index" key, not an "sindex" key.
Actually nothing was broken by this. When the user switches server selections, server_changed() is called. That uses JavaScript and selectedIndex to work out which entry the user selected. selectedIndex is a zero-based number. It matches the zero-based index of servers, which was built from the zero-based indexes of $ras_server array, which was built from data in the config. So by good luck or good management, this array index name typo is not actually a bug.
I noticed it while investigating forum https://forum.pfsense.org/index.php?topic=86263.0


phil-davis:
I do not think this is worth actually "fixing" right now since nothing is actually broken. But it could be included next time there is a genuine reason to update code and version number of the OpenVPN client export package.
I will also see what is the end result of that forum thread above - perhaps there will be some unusual combination of similar settings in 2 servers that is causing only 1 to display, and at that time there might be useful code changes.

PR #784, opened 2015-01-10
Remove the correct snort.sh on deinstall

phil-davis:
Forum report: https://forum.pfsense.org/index.php?topic=86562.0

@bmeeks8 - this just looked wrong, because the end of snort_create_rc() has:
	/* write out snort.sh */
	@file_put_contents("{$rcdir}snort.sh", $snort_sh_text);
	@chmod("{$rcdir}snort.sh", 0755);

which seems to clearly create snort.sh
I do not run snort, so I haven't tested this! Feel free to ignore my pull request, and include this fix in the next set of snort bug fixes and version bump.

### DIFF
--- a/config/snort/snort.inc
+++ b/config/snort/snort.inc
@@ -3094,7 +3094,7 @@ function snort_deinstall() {
 	log_error(gettext("[Snort] Snort package uninstall in progress..."));
 
 	/* Remove our rc.d startup shell script */
-	unlink_if_exists("{$rcdir}snort_pkg.sh");
+	unlink_if_exists("{$rcdir}snort.sh");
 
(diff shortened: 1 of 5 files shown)


bmeeks8:
Good catch.  snort_pkg.sh was the name of alternate shell script I was experimenting with a few months back when trying to reconfigure the package so each running Snort interface would show up as a separate service under SERVICES.  There were complicated startup issues with that scheme I could never overcome with the current package and service interface architecture, so I abandoned it.  However, I forgot to change this line back.

rbgarga:
Please bump package version

phil-davis:
Version bump done.

PR #785, opened 2015-01-13
Updated arpwatch to handle multiple interfaces

CinoNY:
Updated arpwatch to handle multiple interfaces
One report per interface
Multiple child report tabs in report tab
Updated arpwatch.xml versions
Capitalized Services title name

i can't take credit for this but it should be merge in to make use of the arpwatch package

https://forum.pfsense.org/index.php?topic=68401.msg392386#msg392386
https://github.com/parawizard/pfsense-packages/commit/b435f2a31aca8c71954ea19ad3c8ce55d51ba2e9
https://github.com/pfsense/pfsense-packages/pull/614

### DIFF
--- a/config/arpwatch.xml
+++ b/config/arpwatch.xml
@@ -42,18 +42,18 @@
 	<requirements>None</requirements>
 	<faq>Currently there are no FAQ items provided.</faq>
 	<name>arpwatch</name>
-	<version>2.1.a14 pkg v1.1.1</version>
-	<title>arpwatch: Settings</title>
+	<version>2.1.a15 pkg v1.2</version>
+	<title>Arpwatch: Settings</title>
 	<aftersaveredirect>pkg_edit.php?xml=arpwatch.xml&amp;id=0</aftersaveredirect>
 	<menu>
-		<name>arpwatch</name>
+		<name>Arpwatch</name>
 		<tooltiptext>Modify arpwatch settings.</tooltiptext>
 		<section>Services</section>
 		<configfile>arpwatch.xml</configfile>
 		<url>/pkg_edit.php?xml=arpwatch.xml&amp;id=0</url>
 	</menu>
 	<service>
-		<name>arpwatch</name>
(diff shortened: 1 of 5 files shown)


config/arpwatch.xml
```diff
@@ -81,9 +81,12 @@
 	</additional_files_needed>
 	<fields>
 		<field>
-			<fielddescr>Listening Interface</fielddescr>
-			<fieldname>interface</fieldname>
-			<description>Choose the desired listening interface here.</description>
+			<fielddescr>Interfaces</fielddescr>
+			<fieldname>interface_array</fieldname>
```
rbgarga: You renamed field from interface to interface_array, won't it break old configuration?

CinoNY: It will, you will have to select the interfaces you want to monitor after its installed.
It will look like this now:
		<arpwatch>
			<config>
				<interface_array>lan</interface_array>
				<interface_array>opt1</interface_array>
				<interface_array>opt2</interface_array>
				<enable_email>on</enable_email>
			</config>
		</arpwatch>

rbgarga: It would be nice if the interface that was already set by user before upgrade stay set after

CinoNY: I looked over the code and to change this field back, a new variable would have to be create for $interface i think. I believe they would conflict because the original package would only allow you to select 1 interface. I've used arpwatch for a couple of years and every firmware update that will force a package install, i would have to re-save the interface before the interface would start

marcelloc:
If the field with only one interface selected looks the same on xml as previous version then I guest there is no need to rename it.

rbgarga:
Can you please rebase your fork? This PR cannot be automatically merged.

CinoNY:
Not sure out to rebase via the web or windows client... Thought I was able to from the windows client but it didn't work. I'll see what I can do, if I can't; I'll cancel this PR and create a new one

PR #789, opened 2015-01-16
Add new X-Forward Mode option to GUI

CinoNY:
removed 'disable_xforward' field and replaced it with 'xforward_mode' since this a re-write of the GUI option
adding 5 forward_for options per
http://www.squid-cache.org/Versions/v3/3.4/cfgman/forwarded_for.html

https://forum.pfsense.org/index.php?topic=85280.msg476632#msg476632


rbgarga:
@marcelloc can you check this out please?

marcelloc:
Looks fine.

PR #793, opened 2015-01-24
Remove conflicting variable $g in mailreports' mail_report_generate_graph()

brycied00d:
Fixes #4281

### DIFF
--- a/pkg_config.10.xml
+++ b/pkg_config.10.xml
@@ -1192,7 +1192,7 @@
 		<name>mailreport</name>
 		<descr>Allows you to setup periodic e-mail reports containing command output, log file contents, and RRD graphs.</descr>
 		<category>Network Management</category>
-		<version>2.1</version>
+		<version>2.2</version>
 		<status>Stable</status>
 		<required_version>2.2</required_version>
 		<config_file>https://packages.pfsense.org/packages/config/mailreport/mailreport.xml</config_file>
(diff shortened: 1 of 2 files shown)


pkg_config.10.xml
```diff
@@ -1192,7 +1192,7 @@
 		<name>mailreport</name>
 		<descr>Allows you to setup periodic e-mail reports containing command output, log file contents, and RRD graphs.</descr>
 		<category>Network Management</category>
-		<version>2.1</version>
+		<version>2.2</version>
```
phil-davis: The version number is also (unfortunately) stored in a few files. So these others should also be updated:
https://github.com/pfsense/pfsense-packages/blob/master/config/mailreport/mailreport.xml - some central package information, but I think it is not really actually used.
https://github.com/pfsense/pfsense-packages/blob/master/pkg_config.8.xml - 32bit package list for FreeBSD 8 systems (pfSense 2.*.*)
https://github.com/pfsense/pfsense-packages/blob/master/pkg_config.8.xml.amd64 - 64 bit package list for FreeBSD systems (pfSense 2.*.*)
The abpove 2 files were combined into 1 file covering both 32 and 64-bit for FreeBSD 10.*-based systems (pfSense 2.2), which you have already edited here.
It should gradually get simpler as the old pfSense versions and thus files drop away from support!

jim-p:
I merged this a few days ago, should be in the latest version.

PR #794, opened 2015-01-25
Fix 4282 vnstat symlinks for PBI compatibility fail to install

brycied00d:
The package installs a symlink in an attempt to redirect the /usr/local/etc/vnstat.conf into the PBI directory for the binary to use, however this symlink will fail to overwrite the vnstat.conf that the package included. I suspect this symlink gets destroyed anyways when the file is overwritten by the configuration page. Therefore, the solution is to reinstall the symlink as a matter of course in writing the configuration file.

Fixes #4282

### DIFF
--- a/pkg_config.10.xml
+++ b/pkg_config.10.xml
@@ -591,7 +591,7 @@
 		<build_pbi>
 			<port>net/vnstat</port>
 		</build_pbi>
-		<version>1.11_6,2</version>
+		<version>1.11_6,3</version>
 		<status>Stable</status>
 		<required_version>2.2</required_version>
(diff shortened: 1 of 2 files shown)


pkg_config.10.xml
```diff
@@ -591,7 +591,7 @@
 		<build_pbi>
 			<port>net/vnstat</port>
 		</build_pbi>
-		<version>1.11_6,2</version>
+		<version>1.11_6,3</version>
```
phil-davis: The version number will also need updating in the other older pkg_config etc files to match.

jim-p:
I fixed this (and other misc bugs) in an alternate way, commit is pending, will show up on github in the next 15-30 mins.

PR #795, opened 2015-01-25
Fix #4285 lcdproc package is PBI-ignorant

brycied00d:
Symlink the configuration into the appropriate PBI root so LCDd can find it.

Fixes #4285

### DIFF
--- a/pkg_config.10.xml
+++ b/pkg_config.10.xml
@@ -928,7 +928,7 @@
 		<descr>LCD display driver</descr>
 		<website>http://www.lcdproc.org/</website>
 		<category>Utility</category>
-		<version>lcdproc-0.5.7_1 pkg.v.1.0.1</version>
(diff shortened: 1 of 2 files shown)


pkg_config.10.xml
```diff
@@ -928,7 +928,7 @@
 		<descr>LCD display driver</descr>
 		<website>http://www.lcdproc.org/</website>
 		<category>Utility</category>
-		<version>lcdproc-0.5.7_1 pkg.v.1.0.1</version>
+		<version>lcdproc-0.5.7_1 pkg.v.1.0.2</version>
```
phil-davis: Version will also need updating in the older pkg_config etc files to keep it in sync.

jim-p:
I committed a different fix, which has been confirmed working. Try the current version.

PR #803, opened 2015-02-02
Fixed phpservice status not being able to extract status from ps.

tuyan:
This fixes the Status/Services display of the phpservice package.
This should apply to 2.1.5 as well.

### DIFF
--- a/config/phpservice/phpservice.xml
+++ b/config/phpservice/phpservice.xml
@@ -56,6 +56,9 @@
 		<rcfile>phpservice.sh</rcfile>
 		<executable>phpservice</executable>
 		<description>PHP run from a command line as a service.</description>    
+		<custom_php_service_status_command>
+			exec("/bin/pgrep -f phpservice");
+		</custom_php_service_status_command>
 	</service>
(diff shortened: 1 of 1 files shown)


config/phpservice/phpservice.xml
```diff
@@ -56,6 +56,9 @@
 		<rcfile>phpservice.sh</rcfile>
 		<executable>phpservice</executable>
 		<description>PHP run from a command line as a service.</description>    
+		<custom_php_service_status_command>
+			exec("/bin/pgrep -f phpservice");
```
rbgarga: You can consider using is_process_running(), from util.inc here

tuyan:
Hi Renato, 

Unfortunately is_process_running() won't work for this one. 
ps ax |grep phpservice
45508  -  S       0:00.49 /usr/local/bin/php /usr/local/pkg/phpservice.php
68876  0  S+      0:00.01 grep phpservice

pgrep -anx phpservice does not return anything.

rbgarga:
Please bump package version on pkg_config* files

PR #807, opened 2015-02-03
zabbix2 - update packages - second step

dbaio:
A small change for not break users on pfSense 2.1.
Bump version.

### DIFF
--- a/config/zabbix2/zabbix2.inc
+++ b/config/zabbix2/zabbix2.inc
@@ -2,10 +2,10 @@
 /* $Id$ */
 /* ========================================================================== */
 /*
-    zabbix2-proxy.inc
+    zabbix2.inc
     part of the Zabbix package for pfSense
     Copyright (C) 2013 Danilo G. Baio
-	Copyright (C) 2013 Marcello Coutinho
+    Copyright (C) 2013 Marcello Coutinho
 
     All rights reserved.            
 			                                                                  */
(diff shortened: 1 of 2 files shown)


dbaio:
@maarcis There is no problem for me to support both releases (2.2 LTS and 2.4). 

@rbgarga @jim-p  What do you think?

If there is no problem I can send the new package Zabbix 2.2 LTS (Agent/Proxy).

cbuechler:
@dbaio If you're willing to support both, given the dual-versioned nature of Zabbix as @maarcis described above, it's fine to have both. I don't see any reason not to. @rbgarga is out on vacation this week. Unless @jim-p has an objection, go for it.

jim-p:
If the XML/package code is the same and only the binaries are different, then it should be fine to have both so long as both still exist in the FreeBSD ports tree. We've had two variants before at times in the past during this sort of transition.

maarcis:
Yes, both versions have packages in FreeBSD ports. But I should add that multiple versions is not planned to be transitional thing, upstream is actually planning to have next LTS release 3.0 somewhere around May[1]. Version compatibility with Zabbix Server, however, is mostly problem only for Zabbix Proxies[2].

If You would plan to make Zabbix Proxy and Agent continuously available on pfSense and compatible with current stable and LTS Zabbix versions, You would probably end up having 2 versions - one for LTS and one for stable and occasionally introduce 3rd package (either LTS or stable) during transition time, before it pushes out previous package of corresponding version.

[1] http://www.zabbix.com/life_cycle_and_release_policy.php
[2] https://www.zabbix.com/documentation/2.2/manual/appendix/compatibility

jim-p:
yes, I'm aware how they run but it was a poor choice of words on my part.

I pushed a commit just now to have both 2.2 and 2.4 packages. It should show up in 15-30 mins.

dbaio:
Thank you for the feedback @cbuechler @jim-p 

I was thinking in keeping the packages as they are right now and create new ones "Zabbix Agent LTS" and "Zabbix Proxy LTS" with another path of config files.

And the the Packages "Zabbix 2 Agent" and "Zabbix 2 Proxy" continue with Zabbix standard releases.

With the last commit it will break the installation and the 2.2 packages won't work without modifications.

Thanks.

dbaio:
New proposal https://github.com/pfsense/pfsense-packages/pull/821

PR #812, opened 2015-02-04
pfBlockerNG - Update MaxMind CC Archive

BBcan177:
Add new MaxMind "Anonymous Proxy and Satellite Providers" data.


rbgarga:
Please bump package version so users can see a new version is available

BBcan177:
This is only needed for new installations. Existing installs don't need to upgrade. This MaxMind archive is only required if the Users attempt to download from the MaxMind Web site fails, so a version bump is not required.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    From: Renato BotelhoSent: Monday, February 9, 2015 6:30 AMTo: pfsense/pfsense-packagesReply To: pfsense/pfsense-packagesCc: BBcan177Subject: Re: [pfsense-packages] pfBlockerNG - Update MaxMind CC Archive (#812)Please bump package version so users can see a new version is available

—Reply to this email directly or view it on GitHub.

BBcan177:
@rbgarga. Do you have any other concerns with this PR?

rbgarga:
no, will merge it

PR #817, opened 2015-02-10
Respect SMTP notifications disable in mail reports

phil-davis:
At the moment, mail reports package sends email even if "Disable SMTP Notifications" is checked.
Add the equivalent checks in mail_reports.inc mail_report_send() as are in notices.inc notify_via_smtp() and send_smtp_message()
That will make it respect the System:Advanced:Notifications settings in the same way.

### DIFF
--- a/config/mailreport/mail_reports.inc
+++ b/config/mailreport/mail_reports.inc
@@ -185,12 +185,18 @@ function set_mail_report_cron_jobs($a_mailreports) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


phil-davis:
@JimP - I did not bump the version again. You can decide if you want another version bump, or just let all these changes in quick succession become part of Mail Report 2.2

jim-p:
This is done on purpose (the lack of check) -- people want reports but not notifications.

PR #821, opened 2015-02-18
Zabbix - New Packages

dbaio:
The ideia is keep the packages as they were before and create new ones "Zabbix Agent LTS" and "Zabbix Proxy LTS" with another path of config files.

Proposed:
- Rename packages back.
- Alter package description informing the difference between standard releases and LTS releases.
- Create structure for new packages (LTS) and take out lines of code for keeping compatibility with earlier pfSense versions.


dbaio:
ok @jim-p 
done.
thanks.

dbaio:
done. thank you again @jim-p

PR #823, opened 2015-02-18
vhost: Corrected PHP bin-path for  SSL configuration

dariomas:
"bin-path" should be "/usr/local/bin/php"
(same as 7232161e99d60256c1a4ee94ef800f6d4f39764 )


jim-p:
I merged this in, but at least on 2.2 this should probably all be changed to use sockets like lighttpd does for the GUI (see /var/etc/lighty-webConfigurator.conf on a running 2.2 system)

PR #824, opened 2015-02-18
vhost: Added ['custom_php_service_status_command']

dariomas:
Added code to check if the process is running for any vhost lighthttpd configuration.
In this way it's possible to show the service status on each vhost.


nemesisdesign:
:+1:

PR #826, opened 2015-02-21
Validate widget_snort_display_lines

phil-davis:
Currently if you open the snort widget settings and just press "save" with a blank field, it sets the parameter to blank. Subsequently that causes problems trying to fetch the last "" number of alerts, and actually the little "tool" icon never reappears, so you cannot fix it from the dashboard.
1) Setting the number of entries to 0 seems silly, so limit it to a minimum of 1 alert.
2) If the user blanks the box, then unset widget_snort_display_lines to make sure things will always revert to the default.
3) Whatever else the user types in - "abc", "-999", "23xyz99" - try to interpret it as an int (with intval()), then do not let it be less than 1 (not negative or zero).
This forum entry made me touch it and "break" my widget also: https://forum.pfsense.org/index.php?topic=89161.0

### DIFF
--- a/config/snort/snort_alerts.widget.php
+++ b/config/snort/snort_alerts.widget.php
@@ -46,7 +46,7 @@ $alertColClass = "listMRr";
 
 /* check if Snort widget alert display lines value is set */
 $snort_nentries = $config['widgets']['widget_snort_display_lines'];
-if (!isset($snort_nentries) || $snort_nentries < 0)
+if (!isset($snort_nentries) || $snort_nentries <= 0)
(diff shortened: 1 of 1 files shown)


phil-davis:
I did not bump the snort package version. This does not fix or enhance any "real" in snort, so probably not worth flagging it up and thus causing many people to re-install their snort package. When someone has another "real" fix/change for snort then this could be committed at the same time and the package version bumped.

bmeeks8:
Thanks Phil for this patch. I am OK with it being merged whenever one of the pfSense developers can do so.

Bill

PR #829, opened 2015-03-04
Use stable release of check_mk_agent (v1.2.6)

ApolloDS:
HEAD Version is considered BETA, Stable Release is v1.2.6.
See here: http://mathias-kettner.de/check_mk_download.html

### DIFF
--- a/config/checkmk-agent/checkmk.inc
+++ b/config/checkmk-agent/checkmk.inc
@@ -39,7 +39,7 @@ define('ETC_RC_CONF','/etc/rc.conf.local');
 function checkmk_install() {
 	// Download latest check_mk version from head repo
 	$checkmk_bin="/usr/local/bin/check_mk_agent";
-	mwexec("fetch -o {$checkmk_bin} 'http://git.mathias-kettner.de/git/?p=check_mk.git;a=blob_plain;f=agents/check_mk_agent.freebsd;hb=HEAD'");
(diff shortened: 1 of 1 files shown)


ApolloDS:
Please see the comments here: https://redmine.pfsense.org/issues/4084

hjjg:
As of pfSense 2.2-RELEASE the agent isn't working anymore. I did some research but I don't have enough time to provide a full fix. This is what I've done so far:

1. Manually change checkmk.inc to use the latest working agent so you don't need bash anymore and the right inetd.conf in /var/etc
see https://github.com/hjjg/pfsense-packages/commit/5c2ad6e4608287d4fa424a0419903b1954f9fb87#diff-9cc8750071f9ecd8be6b9227145c3375

2. Kill the inetd and start it with 
/usr/sbin/inetd -wW -R 0 /var/etc/inetd.conf
instead of
/usr/sbin/inetd -wW -a 127.0.0.1 -R 0 /var/etc/inetd.conf

The checkmk.inc tries to use /etc/rc.d/inetd to restart the service, which is not working anymore. pkill -HUP inetd would be a quick fix ;)

lunetics:
Any update on this issue?

hjjg:
More work needs to be done here. The handling of inetd is far from optimal. Someone of the core team should take a look at this.

PR #831, opened 2015-03-10
General freeradius2 cleanup and fix for #4346

k-paulius:
Removed <service> and <additional_files_needed> from files other than freeradius.xml, since that is the main package config file where these things need to be defined.

Kill radiusd processes during deinstall. RC script fails to stop radiusd process because it calls script from freeradius package and that package is uninstalled before RC script is called. #4346

### DIFF
--- a/config/freeradius2/freeradius.inc
+++ b/config/freeradius2/freeradius.inc
@@ -76,6 +76,7 @@ if ($pfs_version == "2.2") {
 	}
 	
 function freeradius_deinstall_command() {
+	exec("killall -9 radiusd");
 	return;
 }
 
--- a/config/freeradius2/freeradius.inc
+++ b/config/freeradius2/freeradius.inc
@@ -76,7 +76,15 @@ if ($pfs_version == "2.2") {
 	}
 	
 function freeradius_deinstall_command() {
-	exec("killall -9 radiusd");
+	$pidFile = "/var/run/radiusd.pid";
+	$i = 0;
+
+	while (isvalidpid($pidFile) && $i < 3) {
+		$sig = ($i == 2 ? SIGKILL : SIGTERM);
+		sigkillbypid($pidFile, $sig);
+		sleep(1);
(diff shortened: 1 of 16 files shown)


config/freeradius2/freeradius.inc
```diff
@@ -73,6 +73,7 @@ if ($pfs_version == "2.1") {
 	}
 	
 function freeradius_deinstall_command() {
+	exec("killall -9 radiusd");
```
rbgarga: Is it really necessary? Isn't there a soft way to do it?

k-paulius:
Renato, I have modified process termination code to send SIGTERM first and only if that fails send SIGKILL. Had to rebase this in order to bump package version.

config/freeradius2/freeradiusauthorizedmacs.xml
```diff
@@ -108,61 +101,6 @@
...
-	<additional_files_needed>
-		<prefix>/usr/local/pkg/</prefix>
-		<chmod>0755</chmod>
-		<item>https://packages.pfsense.org/packages/config/freeradius2/freeradiusmodulesldap.xml</item>
-	</additional_files_needed>
-	<additional_files_needed>
-		<prefix>/usr/local/pkg/</prefix>
-		<chmod>0755</chmod>
-		<item>https://packages.pfsense.org/packages/config/freeradius2/freeradiusauthorizedmacs.xml</item>
-	</additional_files_needed>
```
rbgarga: Are you sure all those files are not being used anymore?

k-paulius: Renato, these files are being used, but they are defined in "master" freeradius.xml file. Somebody must have used that file as a base for this file and left <additional_files_needed> tags in it.

k-paulius:
I have managed to bork the rebase of this branch, so I created new pull request #898 for these changes.

PR #833, opened 2015-03-13
Update apcupsd_mail.php

chrisriteco:
In Pfsense version 2.2 the variable/key $config['notifications']['smtp']['ssl'] is set with an empty string value when the parameter 'Enable SMTP over SSL/TLS' is checked in the webgui (and the key not exists when the checkbox is off). The check in this file is for the value 'checked', but is not working and ever happening, thats why I modify this to get it work with Gmail.

### DIFF
--- a/config/apcupsd/apcupsd_mail.php
+++ b/config/apcupsd/apcupsd_mail.php
@@ -61,7 +61,7 @@ $mail = new PHPMailer();
 $mail->IsSMTP();
 $mail->Host = $config['notifications']['smtp']['ipaddress'];
 
-if ($config['notifications']['smtp']['ssl'] == "checked")
+if (isset($config['notifications']['smtp']['ssl']))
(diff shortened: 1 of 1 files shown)


rbgarga:

chrisriteco:
Hi rbgarga, can you help me how to do this, I have no clue. Thx!

dbaio:
Hi @chrisriteco, For bump, see this commit for a clue: https://github.com/dbaio/pfsense-packages/commit/0677ffb88181f442ad9c28be3e776d82eeb1594d

chrisriteco:
@rbgarga I made a new Pull request with the version corrections, hope this way is correct. Thx for the help.

dbaio:
@chrisriteco  on pfSense 2.2, email notifications are not working in this moment.
Check this out, please: https://forum.pfsense.org/index.php?topic=89901.0

PR #838, opened 2015-03-19
Update squid.inc

nliaudat:
Change squid_clamav icap_service from ipv6 to ipv4 for default configuration (or add a choice). 
IPV6 could be disabled for some reason.
Regards

### DIFF
--- a/config/squid3/34/squid.inc
+++ b/config/squid3/34/squid.inc
@@ -1361,9 +1361,9 @@ icap_client_username_header X-Authenticated-User
 icap_preview_enable on
 icap_preview_size 1024
 
-icap_service service_avi_req reqmod_precache icap://[::1]:1344/squid_clamav bypass=off
+icap_service service_avi_req reqmod_precache icap://localhost:1344/squid_clamav bypass=off
 adaptation_access service_avi_req allow all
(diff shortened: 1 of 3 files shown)


marcelloc:
Did you tested it? While I was fixing icap for squid 3.4 It was by default listening on ipv6. As pfsense does use ipv6, it's working fine on my machines.

nliaudat:
Hi, I've tested with ipv4 and it's working fine on my 3 Pfsense 2.2.0.

CinoNY:
@nliaudat Are you 100% that its connecting via IPv4 and not IPv6? I use 'localhost' and it connects via IPv6 and not IPv4.. To confirm, checked the sockets that are opened on your test boxes. You should only see c-icap only listening to IPv6 loopback. When I was testing a few months ago, c-icap could only be compiled for IPv6 or IPv4, but not both (doesn't make sense)

@marcelloc while we were testing a few months back. I would used 'localhost' and didn't have any issues but it was still connecting via IPv6 and not IPv4 since icap was compiled for IPv6.. Using 'localhost' IMHO is the cleaner approach

nliaudat:
@CinoNY You are right, using 'localhost' listen only IPV6. I've checked the opened socket as you said. 
Regards

rbgarga:

rbgarga:
According @marcelloc this should be closed.

PR #840, opened 2015-03-24
Update sqstat.class.php

ReinaldoFeitosa:
Fix erro in Squid 3.2 or later
Field "peer:" is changed to "remote:" and field "me:" changed to "local:"
Cause:
Warning: gethostbyaddr(): Address is not a valid IPv4 or IPv6 address in /usr/local/www/sqstat/sqstat.class.php on line 426

Fix Bug #3936

### DIFF
--- a/config/lightsquid/sqstat.class.php
+++ b/config/lightsquid/sqstat.class.php
@@ -427,7 +427,6 @@ class squidstat{
             else
             # i use ip2long() to make ip sorting work correctly
             if ($resolveip) {
-				echo $ip;
                 $hostname = gethostbyaddr($ip);
                 if ($hostname == $ip)
                      $ip = ip2long($ip); # resolve failed. use (ip2long) key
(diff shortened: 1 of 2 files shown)


CinoNY:
@ReinaldoFeitosa thanks for fixing this.. Only issue I see is if you connection via IPv6. Which really isn't a big deal, not sure if anyone else connects to squid using IPv6 instead of IPv4.

config/lightsquid/sqstat.class.php
```diff
@@ -423,6 +427,7 @@ function parseRequest($data, $group_by = 'host', $resolveip = false) {    	$pars
             else
             # i use ip2long() to make ip sorting work correctly
             if ($resolveip) {
+				echo $ip;
```
rbgarga: Is this echo supposed to be here? Sounds like a debug leftover

ReinaldoFeitosa: sorry, forgot to remove.
Thank U

rbgarga:
In order to have your patches accepted, you must sign Individual or Corporate CLA as described at https://www.pfsense.org/about-pfsense/#legal

Also, please bump package version on pkg_config xml files

ReinaldoFeitosa:
How to bump package?

PR #841, opened 2015-03-25
apcupsd - improve status page

dbaio:
There are some issues with APCupsd status page by misunderstandings.

- add an option for chose host parameter on apcaccess command.
This options is enabled only on pfSense >= 2.2 (APCupsd 3.4.12)
- add a verification if NIS server is running (apcaccess needs to connect on a NIS server, localhost or remote).
- bump version


config/apcupsd/apcupsd_status.php
```diff
@@ -28,6 +28,7 @@
 */
 
 require("guiconfig.inc");
+require_once("/usr/local/pkg/apcupsd.inc");
```
rbgarga: /usr/local/pkg is on php include path, you can drop it here and just use filename

dbaio: ok @rbgarga, done. thanks.

PR #842, opened 2015-03-27
APCd mail

chrisriteco:
In Pfsense version 2.2 the variable/key $config['notifications']['smtp']['ssl'] is set with an empty string value when the parameter 'Enable SMTP over SSL/TLS' is checked in the webgui (and the key not exists when the checkbox is off). The check in this file is for the value 'checked', but is not working and ever happening, thats why I modify this to get it work with Gmail.

### DIFF
--- a/config/apcupsd/apcupsd_mail.php
+++ b/config/apcupsd/apcupsd_mail.php
@@ -61,7 +61,7 @@ $mail = new PHPMailer();
 $mail->IsSMTP();
 $mail->Host = $config['notifications']['smtp']['ipaddress'];
 
-if ($config['notifications']['smtp']['ssl'] == "checked")
+if (isset($config['notifications']['smtp']['ssl']))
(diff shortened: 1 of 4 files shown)


gogglespisano:
apcupsd_mail.php should also have

if (isset($config['notifications']['smtp']['tls']))
	$mail->SMTPSecure =  "tls";

gogglespisano:
mail_reports.inc uses a backwards compatible test

if ((isset($config['notifications']['smtp']['ssl']) && $config['notifications']['smtp']['ssl'] != "unchecked") || $config['notifications']['smtp']['ssl'] == "checked")
		$mail->SMTPSecure =  "ssl";

if ((isset($config['notifications']['smtp']['tls']) && $config['notifications']['smtp']['tls'] != "unchecked") || $config['notifications']['smtp']['tls'] == "checked")
		$mail->SMTPSecure =  "tls";

PR #843, opened 2015-04-01
Fix IPv6 Reverse Zone configuration. Bug#4553

madforic:
This fix was written by cuteredstorm 19 Nov 2013.
https://github.com/cuteredstorm/pfsense-packages/commit/2b4e8084a92c8e9936f1b2fdca8272d29217c20a

But it was never been included into BIND package of pfsense. If you fix this manually you loose all your updates after BIND package upgrade from pfsense GUI.
So, i decided to commit this fix once again.


gogglespisano:
I've included this change in my pull request https://github.com/pfsense/pfsense-packages/pull/845

PR #844, opened 2015-04-03
correction of portrange checking

dneuhaeuser:
configured port needs checking against sysctl parameter "net.inet.ip.portrange.reservedhigh" instead of "net.inet.ip.portrange.first".

### DIFF
--- a/config/squid3/34/squid.inc
+++ b/config/squid3/34/squid.inc
@@ -642,14 +642,14 @@ function squid_validate_reverse($post, &$input_errors) {
 	preg_match("/(\d+)/",`sysctl net.inet.ip.portrange.reservedhigh`,$portrange);
 	if (!empty($port) && !is_port($port))
 		$input_errors[] = 'The field \'reverse HTTP port\' must contain a valid port number';
-	if (!empty($port) && is_port($port) && $port < $portrange[1]){
+	if (!empty($port) && is_port($port) && $port <= $portrange[1]){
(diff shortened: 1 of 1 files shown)


phil-davis:
On pfSense 2.1.5 and 2.2.2 I get:
net.inet.ip.portrange.reservedhigh: 1023
net.inet.ip.portrange.first: 1024
So it seems to me it does not really matter which is used, as long as the comparison is correct.
Using net.inet.ip.portrange.first the current "<" test is correct (anything <1024 is a bad idea), but the error text "must contain a port number higher than 1024" is slightly wrong - the port number could also be equal to 1024.
Using net.inet.ip.portrange.reservedhigh the error text turns out right - "must contain a port number higher than 1023). But the test needs to be <= otherwise the current code here will allow the user to put port 1023.

dneuhaeuser:
you're right that both sysctl values are the same in pfSense 2.1.x and 2.2.x.
BUT: apparently the underlying freebsd in pfSense 2.2.x behaves differently.
processes without root-rights cannot listen on ports <= reservedhigh value anymore!

so the value that needs to be adjusted to do this is: "net.inet.ip.portrange.reservedhigh"
changing "net.inet.ip.portrange.first" does not help at all.

and you're right: the comparison needs to be <=.

example:
to let squid reverse proxy listen to port 443
system tunable "net.inet.ip.portrange.reservedhigh" needs to be set to 442.

rbgarga:
In order to have your patches accepted, you must sign Individual or Corporate CLA as described at https://www.pfsense.org/about-pfsense/#legal

Also please bump package version on pkg_config.xml files so users can see there is an update available

dneuhaeuser:
ok, both done!

PR #845, opened 2015-04-03
BIND package feature additions and code clean up

gogglespisano:
Backward compatible with existing config structure
Add TCP/IP version multi-select list to add -4 or -6 or none to named command line
Add custom global settings field for named.conf (eg. key and server clauses)
Option for IPV6 reverse zones (Bug #4553, Fix from pull request 843)
Change zone view from single multiple values to add zone to more than one view
Add checkbox and text field to use update-policy, overriding/replacing allow-update
Ensure single trailing semi-colon on forwarders and update-policy
Disable/enable more zone fields on zone type change
Don’t create folders for forward zones
Save empty resulting config for slave zones when no DB is found (fixes incorrect/stale zone displayed)
Save empty resulting config for forwarder zones (fixes incorrect/stale zone displayed)
Improve indentation in named.conf

Backward compatible unless conflicting custom statements added
Include allow-query statement in slave and redirect zones

Code cleanup to reflect coding style guidelines and consistency
Tabs only, always uses braces, spacing
C/C++ style comments
Use array_key_exists instead of key_exists
$config[] keys are always enclosed in quotes
Initialize variables when first use is concatenate .=

Bump package number

### DIFF
--- a/config/bind/bind.inc
+++ b/config/bind/bind.inc
@@ -36,108 +36,132 @@ require_once('util.inc');
 require_once('pfsense-utils.inc');
 require_once('pkg-utils.inc');
 require_once('service-utils.inc');
-if(!function_exists("filter_configure")) 
+if (!function_exists("filter_configure")) {
 	require_once("filter.inc");
-
-$pf_version=substr(trim(file_get_contents("/etc/version")),0,3);
-if ($pf_version > 2.0)
-	define('BIND_LOCALBASE', '/usr/pbi/bind-' . php_uname("m"));
-else
-	define('BIND_LOCALBASE','/usr/local');
-
-define('CHROOT_LOCALBASE','/cf/named');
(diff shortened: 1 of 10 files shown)


rbgarga:
Could you please split it in small commits? It will make the review process much easier. Also, please do not mix code style fixes with functional changes in the same commit.

gogglespisano:
What is the best/preferred procedure for submitting multiple incremental or related commits?

pull request 1 = code commit 1 + code commit 2 + package bump commit
or
pull request 1 = code commit 1
pull request 2 = code commit 2
pull request 3 = package bump commit
or
pull request 1 = code commit 1 + package bump commit
pull request 2 = code commit 2 + package bump commit

or some other variant?

rbgarga:
first method is the best IMO.

PR #849, opened 2015-04-07
Apcupsd - Bug 4650 - Support TLS email notifications and update email no...

gogglespisano:
Apcupsd - Fix #4650 - Support TLS email notifications and update email notification flag tests

Bump package number to 0.3.5


rbgarga:
@dbaio is checking it and will send a feedback

dbaio:
it seems ok @rbgarga   is similar with https://github.com/pfsense/pfsense-packages/blob/master/config/mailreport/mail_reports.inc#L201-L205

@gogglespisano just a note, this feature (phpMailer and APCUpsd) on pfSense 2.2.X is not working.
See APCupsd package status for pfSense:
https://forum.pfsense.org/index.php?topic=89901.0
and thanks for this PR

PR #852, opened 2015-04-07
correction of path to squid binary

dneuhaeuser:
corrected the path to the squid binary
should be SQUID_BASE instead of SQUID_LOCALBASE


rbgarga:

dneuhaeuser:
ok, I have done that on pkg_config.10.xml

now I'm unsure whether the versions of "squidGuard" and "squidGuard-squid3" in pkg_config.8.xml(.amd64) should be bumped to an equal version number...

these packages share the same code, don't they?

rbgarga:
Yes, please bump on both pkg_config.8.xml and pkg_config.8.xml.amd64

PR #854, opened 2015-04-09
BIND - Add new features and clean up code

gogglespisano:
Add several backwards compatible features
- Global settings for named.conf
- Support for IPv6 reverse zones
- Support for named command line options -4 and -6
- Support for zone update-policy statement on master zones (breaks backward compatibility if update-policy already in custom zone options, but unlikely since it would have conflicted with allow-update statement)
- Support for multiple views per zone configuration

Improve indentation in generated named.conf file
General code clean up

Bump package version to 0.3.7


config/bind/bind_zones.xml
```diff
@@ -305,6 +314,22 @@
 			<size>03</size>
 		</field>
 		<field>
+			<fielddescr>Enable Update-policy</fielddescr>
+			<fieldname>enable_updatepolicy</fieldname>
+			<description><![CDATA[Enable Update-policy which overrides Allow-update.<br>
+						The update-policy statment replaces the allow-update statement.]]></description>
```
phil-davis: s/statment/statement

gogglespisano:
Fixed a typo in "statment" => "statement"

DeveloperJr:
On this new package, we need these changes on bind.inc file.
182,183c182,183
<               responses-per-second { $rate_limit };
<               log-only { $log_only };
---
>               responses-per-second $rate_limit;
>               log-only $log_only;

gogglespisano:
It's looks like the {} have been there since the feature was added 18 months ago.

I'll fix it.

DeveloperJr:
Thanks Stuart,

Just for your information: everything was running fine until yesterday.
When i did the upgrade from PfSense 2.2.1-Release to 2.2.2 after the 
boot packages were re-installed; i was using 9.9.6P1 pkg 0.3.6 the auto 
re-install process installed the new one package (pkg 0.3.7) and this 
new one is putting curly braces around reponses-per-second and log-only 
options.

I've just checked that package 0.3.6 is not putting the "{}".

Best Regards,

Valmir Filho

gogglespisano:
I see what happened.  I added spaces inside the {}

responses-per-second {$rate_limit};
log-only {$log_only};

to

responses-per-second { $rate_limit };
log-only { $log_only };

which changed the way the variables were interpreted and made the {} literal text.  I thought I was cleaning up named.conf formatting but it actually changed the text output.

So rather than removing the braces, which also works, I'll put them back and remove the spaces to make it consistent with the rest of the code.  I'll also add comments where <<<EOD uses braces so there's no confusion between name.conf content {} and PHP {}.

Thanks for pushing back,
-Stuart

PR #859, opened 2015-04-12
Squidguard fix ACL order problem

dvserg:
If empty ACL list - pkg_edit.inc show error string.


rbgarga:
In order to get patches applied you must sign Individual or Corporate CLA as described at https://www.pfsense.org/about-pfsense/#legal

dvserg:
You will remove all my packages? Why?
  ----- Original Message -----

dvserg:
I already sent my CLA to the coreteam@ , license@ is a mail delivery.

PR #863, opened 2015-04-16
BIND - fix rate-limit option - remove {}

gogglespisano:
The rate-limit option should not have values with {} and look like this:

rate-limit {
    responses-per-second 15;
    log-only yes;
  };


rbgarga:
Please bump package version

gogglespisano:
Sorry, forgot to bump the package - done

PR #865, opened 2015-04-18
Allow open bg pd use with carp and ix

athompso:
I think "depend on" might normally be handled by local-address, but not everyone wants to do it that way.
The other setting, enforce neighbor-as, is required when connecting to an IX that hides its internal ASN.


rbgarga:
Applied a different patch in 2fafdcbadfae5173a3c4dcd45b24955534d087b4

PR #868, opened 2015-04-22
Use last working stable release of check-mk-agent (v1.2.5i7)

8191:
Never versions are not compabile with pfsense since they
depend on bash instead of sh.


rbgarga:
Please bump package version on pkg_config* xml files

doktornotor:
There are also references to this in XML. (This really is an absurd method of development, the upstream guy does not do releases or WTF.)

I'm doing a complete code cleanup ATM.

rbgarga:
I'm closing this one in favor of #932

PR #870, opened 2015-04-28
Updated routed to include passive RIP interfaces

marcusbel:
To be able to select individual interfaces where you do not want to advertise routes.


rbgarga:

cbuechler:
Missing CLA, has merge conflicts, and needs to be updated and submitted for 2.3

PR #871, opened 2015-04-30
Update havp directories for 2.2.2 

ncolunga:
This changes fix HAVP on pfsense 2.2.2 as related on https://forum.pfsense.org/index.php?topic=90706.0

### DIFF
--- a/config/havp/havp.inc
+++ b/config/havp/havp.inc
@@ -76,7 +76,12 @@ define('HVDEF_PROXYPORT',	            '8080');
 define('HVDEF_MAXSCANSIZE',             '5000000'); # [bytes] ! do not enter 0 or big size !
 define('HVDEF_MAXARCSCANSIZE',          '5000000'); # [bytes] ! do not enter 0 or big size !
 define('HVDEF_PID_FILE',                '/var/run/havp.pid');
-define('HVDEF_WORK_DIR',                '/usr/pbi/havp-amd64/local/etc');
+
+if ($pf_version == "2.1" || $pf_version == "2.2")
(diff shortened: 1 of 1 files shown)


BBcan177:
Just to let you know that It looks like Havp will be only available for 2.2.2 and below. 

[Github LINK](https://github.com/pfsense/pfsense-packages/commit/31372afe3ed3fb8691aa726adb75cd56c5b1f290)

phil-davis:
Just to clarify, that change makes it available up to 2.2.999 - effectively 2.2.* - so for all 2.2.* bugfix releases - 2.2.3...

ncolunga:
I'm not sure if it's a good idea to increase package to 2.3.*

config/havp/havp.inc
```diff
@@ -76,7 +76,7 @@ define('HVDEF_PROXYPORT',	            '8080');
 define('HVDEF_MAXSCANSIZE',             '5000000'); # [bytes] ! do not enter 0 or big size !
 define('HVDEF_MAXARCSCANSIZE',          '5000000'); # [bytes] ! do not enter 0 or big size !
 define('HVDEF_PID_FILE',                '/var/run/havp.pid');
-define('HVDEF_WORK_DIR',                '/usr/local/etc/havp');
+define('HVDEF_WORK_DIR',                '/usr/pbi/havp-amd64/local/etc');
```
rbgarga: You need to check pfSense version and also system arch (amd64 or i386) before set it. Take a look at an example at https://github.com/pfsense/pfsense-packages/blob/master/config/snort/snort_defs.inc#L38

ncolunga:
Version and arch checks done.

cbuechler:
Merged, thanks!

PR #872, opened 2015-05-08
Added (default enabled) option to name exported VPN config based on V…

funkypenguin:
…PN description

We use the client-export utility to create VPN configs for our users, and the
default naming convention of <hostname>-<protocol>-<port>-<user> is a little
unfriendly. I modified the web UI and the function which generates the config
to pass a flag (default enabled) which will instead name the VPN

### DIFF
--- a/config/openvpn-client-export/openvpn-client-export.inc
+++ b/config/openvpn-client-export/openvpn-client-export.inc
@@ -44,17 +44,20 @@ $current_openvpn_version_rev = "03";
 
 function openvpn_client_export_install() {
 	global $current_openvpn_version;
-	conf_mount_rw();
-	$tarpath = "/usr/local/pkg/openvpn-client-export-{$current_openvpn_version}.tgz";
-	$phpfile = "vpn_openvpn_export.php";
-	$ovpndir = "/usr/local/share/openvpn";
-	$workdir = "{$ovpndir}/client-export";
 
-	if (!is_dir($workdir))
-		mkdir($workdir, 0777, true);
+	$pfs_version = substr(trim(file_get_contents("/etc/version")),0,3);
+	if ($pfs_version == "2.1" || $pfs_version == "2.2") {
+		conf_mount_rw();
+		$tarpath = "/usr/local/pkg/openvpn-client-export-{$current_openvpn_version}.tgz";
+		$ovpndir = "/usr/local/share/openvpn";
+		$workdir = "{$ovpndir}/client-export";
 
-	exec("/usr/bin/tar zxf {$tarpath} -C {$ovpndir}");
-	conf_mount_ro();
+		if (!is_dir($workdir))
+			mkdir($workdir, 0777, true);
+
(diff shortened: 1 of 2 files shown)


config/openvpn-client-export/openvpn-client-export.inc
```diff
@@ -44,20 +44,17 @@ $current_openvpn_version_rev = "03";
 
 function openvpn_client_export_install() {
 	global $current_openvpn_version;
+	conf_mount_rw();
+	$tarpath = "/usr/local/pkg/openvpn-client-export-{$current_openvpn_version}.tgz";
+	$phpfile = "vpn_openvpn_export.php";
```
phil-davis: I have not looked hard, but why does this line need to be added?
And why do you need to take all this code here out of the $pfs_version tests - maybe so it works for 2.3 and onwards?

funkypenguin: Err... that's actually a mistake. My bad - github newbie here. I think this is because my version of the file is based on what's currently found on a live pfsense box, and it looks as if the git master branch has a newer version. I'll update my branch with an amendment, which (I believe) should update this pull request.

config/openvpn-client-export/openvpn-client-export.inc
```diff
@@ -807,7 +816,7 @@ function openvpn_client_export_sharedkey_config($srvid, $useaddr, $proxy, $zipco
 			}
 			$conf .= "http-proxy {$proxy['ip']} {$proxy['port']} ";
 		}
-		if ($proxy['proxy_type'] == "socks") 
+		if ($proxy['proxy_type'] == "socks")
```
funkypenguin: This seems to be caused by atom/vi stripping blank spaces at the end of the lines. Since this actually improves the code (there are obviously no extra spaces on any other lines, else they would have suffered the same treatment), I suggest letting these minor corrections through as part of the pull.

phil-davis: Yes, that should be fine. The pfSense developer style guide is at https://doc.pfsense.org/index.php/Developer_Style_Guide
It specifies no white space at EOL.
And other stuff that is randomly used in code - like right here there are "if(" and "if (" used at random. "if (" is the standard. I am standardizing a lot of the core PHP code at the moment. Maybe there will be a day to do the same to all the package code :)

rbgarga:

funkypenguin:
I've signed an ICLA as above :)

D

funkycopter:
Noob here - how do I resolve the conflicts reported above?

funkypenguin:
Oops, that was me, above, signed into a different account.

jim-p:
It would be best to redo the original edits against the new version of this package over in https://github.com/pfsense/FreeBSD-ports/tree/devel/security/pfSense-pkg-openvpn-client-export
See also here: https://forum.pfsense.org/index.php?topic=103481.0

PR #873, opened 2015-05-11
Allow multiple destinations in ftp proxy bypass

davidcornish:
In the FTP proxy it is only possible at present to provide a single destination.

This allows multiple destinations to be specified that will bypass the proxy if they are separated by a comma.

They need not all be the same type (one may be an IP address, another an alias for example) and spaces around the commas are ignored.


jim-p:
Use of an alias for that is preferred over allowing multiple inputs.

PR #874, opened 2015-05-13
Update openbgpd_neighbors.xml

jshupe:


rbgarga:
What are you trying to fix/implement with this change?

cbuechler:
prepend self might be handy to have in the GUI, but there's a lot more to it than this. Thus far everyone that's needed it is in a situation where they need to use raw config for a variety of other reasons. Closing this out since it won't work as it stands. 

It is something we'd get merged if it were complete though @jshupe if you want to pursue it further.

jshupe:
We use this patch in production, but... I'll look at into expanding it.

PR #877, opened 2015-05-15
zabbix_lts - fix resync problem

dbaio:
- change exec(killall) for sigkillbyname() function on deinstall.
- drop restart_command from xml.
- insert a double-check if process is running on resync function.

### DIFF
--- a/config/zabbix-agent-lts/zabbix-agent-lts.inc
+++ b/config/zabbix-agent-lts/zabbix-agent-lts.inc
@@ -54,7 +54,7 @@ function php_deinstall_zabbix_agent_lts(){
 
    conf_mount_rw();
 
-	sigkillbyname("zabbix_agentd", "TERM");
+   sigkillbyname("zabbix_agentd", "TERM");
    unlink_if_exists(ZABBIX_AGENT_BASE . "/etc/rc.d/zabbix_agentd_lts.sh");
    unlink_if_exists(ZABBIX_AGENT_BASE . "/etc/zabbix22/zabbix_agentd.conf");
    unlink_if_exists("/var/log/zabbix-agent-lts/zabbix_agentd_lts.log");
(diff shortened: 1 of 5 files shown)


config/zabbix-agent-lts/zabbix-agent-lts.inc
```diff
@@ -257,11 +257,28 @@ EOF;
 						"stop" => "$zagent_stop"
 				)
 		);
-		mwexec("{$zagent_rcfile} restart");	
+
+		// Check if zabbix_agentd is currently running
```
rbgarga: Please use /* */ notation for multiple comment lines

config/zabbix-agent-lts/zabbix-agent-lts.inc
```diff
@@ -257,11 +257,28 @@ EOF;
...
+		}else{
+			mwexec("{$zagent_rcfile} stop");
+			// Check if zabbix_agentd is still running
+			if (is_process_running("zabbix_agentd")) {
+				sigkillbyname("zabbix_agentd", "KILL");
+				sleep(2);
+				unlink_if_exists("/var/run/zabbix-agent-lts/zabbix_agent_lts.pid");
+			}
+			// Start zabbix_agentd
+			mwexec("{$zagent_rcfile} start");
```
rbgarga: Have you checked start_service() at service-utils.inc?

dbaio: yes, but on zabbix_proxy there is a delay for stop the service, and when you do a resync the start command was being executed before it stops completely. (it happens some times)

rbgarga: This dirty game (kill -TERM, check, sleep, kill -KILL, ...) should be done inside rc script instead of at PHP side. It should work as expected since stop_service() waits until rc file is executed.

dbaio:
@rbgarga ok I will do that, also I will close this PR for not send a lot of commits. 
thanks again.

PR #880, opened 2015-05-23
pfBlockerNG v1.09 05-2015-1

BBcan177:
## Changelog:

- Code Style improvements
- Add option to customize the Inbound Ports and Destination Addresses using pfSense Aliases.
- Add function to check if Cron task needs updating. Minimize Config backups.
- Cron time function moved from .php to .inc file
- Cron settings now allow for hour 'interval' customizations.
- Added preliminary code "$pfb['cron_mod']" which will only save the config when changes are found. This will avoid creating config backups when not required. Not fully implemented.
- Remove code to determine pfSense version below 2.2
- Remove code to check for presence of pfBlocker original version.
- Reduce blank lines in pfblockerng.log
- Adjust IP regex for CIDR matching
- Remove '/32' from Lists as pf strips this out anyways
- When parsing IP lists, use regex to find Ranges first before CIDR or single addresses.
- Add Input validation for IPv4/6 Tab inputs
- Improve log text in XMLRPC sync function and add missing quotes around key ['row']
- If continent blocking is used solely, Cron will check for Continent setting changes each time cron is called.
- Improve XML code style. Remove some unneeded Tags.
- Use pkg_edit.php "combinefields" tags to group settings for ease of use.
- IPv4/6 Continent Tables are now side-by-side to make it easier to configure.
- IPv4/6 Aliases can be now be re-ordered.
- Remove reference to "Gold Membership" as everyone thinks I work and collect a commission from pfSense:)
- priv.inc file, Incorrect reference for IPv4/6 tabs to pkg_edit.php when they should be pkg.php
- Complete overhaul of widget. Added "Wrench icon" to configure additional settings.
  Maxfails, Pivot to alerts tab, Sort column, and Widget Rule table Popup.
- Clicking on widget packet count will pivot to the Alerts Tab and open the filter search to show the selected packet alerts.
- Improve the collection of Local IP addresses in the Alerts Tab (function ip_in_pfb_localsub)
- Alerts Tab, remove check for previous pfSense versions
- pfblockerng_diag_dns.php, remove check for previous pfSense versions
- Update Tab, use the new Cron Interval settings to report the next cron run in hr:min:sec remaining.

### DIFF
--- a/config/pfblockerng/pfblockerng_diag_dns.php
+++ b/config/pfblockerng/pfblockerng_diag_dns.php
@@ -56,15 +56,6 @@ foreach ($a_aliases as $a) {
 	$counter++;
 }
 
-# Collect pfSense Version
-$pfs_version = substr(trim(file_get_contents("/etc/version")), 0, 3);
-
-if ($pfs_version > '2.2') {
-	$cmd = '/usr/bin/drill';
-} else {
-	$cmd = '/usr/bin/dig';
-}
-
 if (isset($_POST['create_alias']) && (is_hostname($host) || is_ipaddr($host))) {
 	if ($_POST['override']) {
 		$override = true;
@@ -73,7 +64,7 @@ if (isset($_POST['create_alias']) && (is_hostname($host) || is_ipaddr($host))) {
 	$type = "hostname";
 	if ($resolved) {
 		$resolved = array();
-		exec("{$cmd} {$host_esc} A | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
+		exec("/usr/bin/drill {$host_esc} A | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
 		$isfirst = true;
 		foreach ($resolved as $re) {
 			if ($re <> "") {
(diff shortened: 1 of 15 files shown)


config/pfblockerng/pfblockerng_diag_dns.php
```diff
@@ -73,7 +64,7 @@
 	$type = "hostname";
 	if ($resolved) {
 		$resolved = array();
-		exec("{$cmd} {$host_esc} A | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
+		exec("/usr/bin/drill {$host_esc} A | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
```
rbgarga: It would be nice to replace exec() by a native PHP function, http://php.net/manual/en/function.dns-get-record.php

BBcan177:
First look at this, found that we need to use "htmlspecialchars" instead of the existing "escapeshellarg" which adds quotes around the $host_esc variable.

$host_esch = htmlspecialchars($host, ENT_NOQUOTES);

I don't think we need to add the "DNS_AAAA" type as "gethostbyname" seems to resolve the IPv4 address?

$resolved = dns_get_record($host_esch, DNS_A|DNS_AAAA);
$resolved = $resolved[0]['host'];    <---- So this seems to be fine? IPv6 is in key [1]

Maybe should also add "@" infront of dns_get_record?

(Ultimately, this function is not really that useful for pfBNG.. The main reason for this php file is to add other Lookup Threat Source Links.)
===================================
sample:

host pfsense.org
host_esc pfsense.org
gethostbyname: 208.123.73.69

array(2) {
  [0]=>
  array(5) {
    ["host"]=>
    string(11) "pfsense.org"
    ["class"]=>
    string(2) "IN"
    ["ttl"]=>
    int(1646)
    ["type"]=>
    string(1) "A"
    ["ip"]=>
    string(13) "208.123.73.69"
  }
  [1]=>
  array(5) {
    ["host"]=>
    string(11) "pfsense.org"
    ["class"]=>
    string(2) "IN"
    ["ttl"]=>
    int(3447)
    ["type"]=>
    string(4) "AAAA"
    ["ipv6"]=>
    string(18) "2610:160:11:11::69"
  }
}

Resolved: pfsense.org

----------------------------------------------------------------

This also looks incorrect in the original code:

                       if ($host != $resolved) {
                                $ipaddr = $resolved[0];
                        }

Should be $ipaddr = $resolved[0]['ip'];   ??

------------------------------------------------------------------

Also don't have a solution for this one?

$query_time = exec("/usr/bin/drill {$host_esc} " . escapeshellarg("@" . trim($dns_server)) . " | /usr/bin/grep Query | /usr/bin/cut -d':' -f2"
);

------------------------------------------------------------------

Comments? and I can submit a new commit.... Thanks!

PR #881, opened 2015-05-24
haproxy-devel, new functionality SSL OCSP stapling for haproxy

PiBa-NL:
haproxy-devel, new functionality SSL OCSP stapling for haproxy
-fix server dns resolving in case a CNAME is used.

### DIFF
--- a/config/haproxy-devel/pkg/haproxy.inc
+++ b/config/haproxy-devel/pkg/haproxy.inc
@@ -346,9 +346,11 @@ function haproxy_custom_php_deinstall_command() {
 	$static_output .= "HAProxy, deleting haproxy webgui\n";
 	update_output_window($static_output);
 	exec("rm /usr/local/etc/rc.d/haproxy.sh");
+	exec("rm /etc/rc.haproxy_ocsp.sh");
 	$static_output .= "HAProxy, installing cron job if needed\n";
 	update_output_window($static_output);
 	haproxy_install_cron(false);
+	haproxy_install_cronjob(false, '/etc/rc.haproxy_ocsp.sh');
 	$static_output .= "HAProxy, running haproxy_custom_php_deinstall_command() DONE\n";
 	update_output_window($static_output);
 }
@@ -433,6 +435,31 @@ EOD;
 	fclose($fd);
 	exec("chmod a+rx /usr/local/etc/rc.d/haproxy.sh");
 
+	$haproxy_ocsp = <<<EOD
+#!/usr/local/bin/php -f
+
+<?php
+
+/*
+	Updates haproxy OCSP responses.
+*/
+
+require_once("globals.inc");		
+require_once("functions.inc");
+require_once("haproxy.inc");
+require_once("haproxy_socketinfo.inc");
+haproxy_updateocsp();
+
+?>
+
+EOD;
+	// removing the \r prevents the "No input file specified." error..
+	$haproxy_ocsp = str_replace("\r\n","\n", $haproxy_ocsp);
+	$fd = fopen("/etc/rc.haproxy_ocsp.sh", "w");
+	fwrite($fd, $haproxy_ocsp);
+	fclose($fd);
+	exec("chmod a+rx /etc/rc.haproxy_ocsp.sh");
+
 	$static_output .= "HAProxy, update configuration\n";
 	update_output_window($static_output);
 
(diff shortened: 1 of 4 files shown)


config/haproxy-devel/pkg/haproxy.inc
```diff
@@ -346,9 +346,11 @@ function haproxy_custom_php_deinstall_command() {
 	$static_output .= "HAProxy, deleting haproxy webgui\n";
 	update_output_window($static_output);
 	exec("rm /usr/local/etc/rc.d/haproxy.sh");
+	exec("rm /etc/rc.haproxy_ocsp.sh");
```
rbgarga: Please use @unlink() or unlink_if_exists() instead of opening a shell session

config/haproxy-devel/pkg/haproxy.inc
```diff
@@ -433,6 +435,31 @@ EOD;
...
+
+?>
+
+EOD;
+	// removing the \r prevents the "No input file specified." error..
+	$haproxy_ocsp = str_replace("\r\n","\n", $haproxy_ocsp);
+	$fd = fopen("/etc/rc.haproxy_ocsp.sh", "w");
+	fwrite($fd, $haproxy_ocsp);
+	fclose($fd);
+	exec("chmod a+rx /etc/rc.haproxy_ocsp.sh");
```
rbgarga: Replace it by a chmod() call

config/haproxy-devel/pkg/haproxy_utils.inc
```diff
@@ -48,7 +48,7 @@ class haproxy_utils {
 				if (haproxy_utils::$pf_version < '2.2')
 					exec("/usr/bin/dig   {$host_esc} $type @$dnsserver | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
 				else
-					exec("/usr/bin/drill {$host_esc} $type @$dnsserver | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
+					exec("/usr/bin/drill {$host_esc} $type @$dnsserver | /usr/bin/awk '{ if($1!=\";;\" && $4==\"{$type}\") print $5 }'", $resolved);
```
rbgarga: PHP provides a mechanism to get DNS entries that would simplify a lot this code IMO:  http://php.net/manual/en/function.dns-get-record.php

pkg_config.10.xml
```diff
@@ -184,7 +184,7 @@
 		<build_pbi>
 			<ports_before>security/openssl</ports_before>
 			<custom_name>haproxy-devel</custom_name>
-			<port>net/haproxy</port>
+			<port>net/haproxy-devel</port>
```
rbgarga: Please change <run_depends> accordingly

PiBa-NL: I hope `<run_depends>sbin/haproxy:net/haproxy-devel</run_depends>` is ok? I have commited that below in beaf604.

config/haproxy-devel/pkg/haproxy_utils.inc
```diff
@@ -36,26 +36,29 @@ require_once("config.inc");
...
-				else
-					exec("/usr/bin/drill {$host_esc} $type @$dnsserver | /usr/bin/grep {$host_esc} | /usr/bin/grep -v ';' | /usr/bin/awk '{ print $5 }'", $resolved);
-				foreach($resolved as $item) {
-					$newitem = array();
-					$newitem["typeid"] = $type;
-					$newitem["data"] = $item;
-					$result[] = $newitem;
-				}
+			switch ($type) {
+				case 'A': $recordtypes += DNS_A; break;
```
rbgarga: Please break it in multiple lines as you did in line 54

config/haproxy-devel/pkg/haproxy_utils.inc
```diff
@@ -36,26 +36,29 @@ require_once("config.inc");
...
+				case 'AAAA': $recordtypes += DNS_AAAA; break;
+			}
+		}
+		$dnsresult = dns_get_record($host, $recordtypes);
```
rbgarga: $recordtypes was initialised as 0, would be good to check if it's still 0 and do not call dns_get_record(), or any previous check guarantee it will always have a value?

config/haproxy-devel/pkg/haproxy_utils.inc
```diff
@@ -36,26 +36,29 @@ require_once("config.inc");
...
+		foreach($dnsresult as $item) {
+			$type = $item['type'];
```
rbgarga: Why set a variable if you can use $item['type'] directly?

config/haproxy-devel/pkg/haproxy_utils.inc
PiBa-NL: Basically the calling code always asks A or AAAA.. But your right its nice for future purposes if the function itself handles checks this. Even though dns_get_record itself will happily return nothing when called with 0.. Changing in coming commit.

config/haproxy-devel/pkg/haproxy_utils.inc
PiBa-NL: I use the value twice, and find it slightly better readable this way.. Ill change it anyway.

config/haproxy-devel/pkg/haproxy_utils.inc
PiBa-NL: Ok.

rbgarga:
Does it require a PBI rebuild?

PiBa-NL:
The OCSP functionallity is available in haproxy 1.5. So it is not required for this specific pull-request. I would 'like' to have the 1.6-dev1 version build on pfSense though.. But that may happen at a later time. The changes in this pull-request will be copied to haproxy1_5 package when they have been tested and commented on by some more haproxy users.

PR #882, opened 2015-05-28
haproxy-devel, include lua53 for compilation of haproxy 1.6-dev1

PiBa-NL:
haproxy-devel, include lua53 for compilation of haproxy 1.6-dev1

@rbgarga , please also check openssl version that gets compiled into the pbi today seems to be confused what version it uses when running "haproxy -vv" after installing the pbi it outputs:
Built with OpenSSL version : OpenSSL 1.0.2a 19 Mar 2015
Running on OpenSSL version : OpenSSL 1.0.1m 19 Mar 2015 (VERSIONS DIFFER!)
Should any additional build options be set? Or is it a problem with the buildserver/scripts perhaps?

### DIFF
--- a/pkg_config.10.xml
+++ b/pkg_config.10.xml
@@ -158,7 +158,7 @@
 		<conflicts>haproxy-devel</conflicts>
 		<depends_on_package_pbi>haproxy-devel-1.5.11-##ARCH##.pbi</depends_on_package_pbi>
 		<build_pbi>
-			<ports_before>security/openssl</ports_before>
+			<ports_before>security/openssl lang/lua53</ports_before>
 			<custom_name>haproxy-devel</custom_name>
 			<port>net/haproxy</port>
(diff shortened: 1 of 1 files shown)


rbgarga:
A new PBI was built.

Probably haproxy binary is looking for system's openssl while it should use openssl from inside PBI.

PiBa-NL:
@rbgarga,
Even openssl itself seems to be confused..? seems to me somehow the build process is going wrong:

[2.2.2-RELEASE][root@pfSense.localdomain]/usr/pbi/haproxy-devel-amd64/local/bin: /usr/pbi/haproxy-devel-amd64/local/bin/openssl version
OpenSSL 1.0.2a 19 Mar 2015 (Library: OpenSSL 1.0.1m 19 Mar 2015)

Can you check? Thanks.

PR #887, opened 2015-06-20
haproxy-devel, dont send the root CA cert to clients when using ssl-offloading, its not needed.

PiBa-NL:
haproxy-devel, dont send the root CA cert to clients when using ssl-offloading, its not needed.
-cipher example and HSTS textual addition
-another try to include lua for haproxy1.6dev package binary

### DIFF
--- a/config/haproxy-devel/pkg/haproxy.inc
+++ b/config/haproxy-devel/pkg/haproxy.inc
@@ -957,18 +957,23 @@ function haproxy_write_certificate_crl($filename, $crlid, $append = false) {
 	unset($crl);
 }
 
-function haproxy_write_certificate_fullchain($filename, $certid, $append = false) {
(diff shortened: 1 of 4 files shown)


rbgarga:
For some reason, this commit was already included in #889, closing...

PiBa-NL:
@rbgarga, its because you like version bumps so much, this commit bumps version to 0.25 and then #889 bumped version to 0.26. Which needed to rely on this first commit, as else there would have been a merge conflict if 889 would still have expected to update from 0.24..

Anyway thanks for merging.

rbgarga:
@PiBa-NL It's not a personal like, without a version bump users won't see a new version is available. Probably it happened because other Pull Request's branch born from this one with the first commit applied. Anyway thanks for contributing :)

PR #888, opened 2015-06-20
ntopng, fix mkdir warning while booting, use safe_mkdir

PiBa-NL:
ntopng, fix mkdir warning while booting, use safe_mkdir


rbgarga:
Please bump package version

PR #894, opened 2015-07-09
Minor Stunnel Improvements

djmarcin:
This fixes stunnel start/stop from the GUI by correcting the rcfile in stunnel.xml and adds the ability to create client mode tunnels from the GUI.

Tested with a local package repository and it appears to work properly.

The package still fails to start if the "cert = ..." line is present, which I think is due to an outdated package.  Installing stunnel 5.19 with "pkg install stunnel" fixes that issue, so I think the pbi needs to be updated as well, but I don't have access to the pfsense-tools repo to hack on that.


rbgarga:
Please bump package version on pkg_config*.xml files

djmarcin:
Bumped version number in pkg_config.10.xml and tested that it shows an upgrade available.  Not sure if I should be doing anything in pkg_config.8 because those look much older.

I tried to bring up pfsense 2.0 in a virtualbox to install these packages but was having trouble.  If you want me to further on old versions I'll give it another shot.

PR #899, opened 2015-07-13
tinydns - fix Bug #2720

doktornotor:
Add patch by Yonas Yanfa; cannot use file_exists with wildcard.


cbuechler:
merged, thanks

PR #903, opened 2015-07-23
System patches: Add status column and other GUI improvement

stilez:
For ease of use:

1) Add a single "Status" column
2) Merge "Apply" and "Revert" into a single column (they are mutually exclusive reverses of each other so it flows better)
3) Fix logic at old lines 225-228: at present if location and patch both blank, it will offer "fetch", but it can't fetch a patch if a location isn't held, so this is incorrect. Need to check empty(location) *first* and then, *if* a location is held, use empty(patch) to decide whether the GUI text is "fetch" or "refetch". (Fixed code is @ lines 241-242)


doktornotor:
Please, review this ticket: https://redmine.pfsense.org/issues/4895

Thanks. ;)

PR #904, opened 2015-07-23
Bug #4583 - Squid3 breaking /etc/inc/captiveportal.inc

doktornotor:
Remove the broken patching nonsense that damages core OS files; reference: https://forum.pfsense.org/index.php?topic=91435.0 and many other threads on the forum.

### DIFF
--- a/config/squid3/34/squid.inc
+++ b/config/squid3/34/squid.inc
@@ -2130,48 +2130,7 @@ function squid_generate_rules($type) {
 	global $config;
 
 	$squid_conf = $config['installedpackages']['squid']['config'][0];
-	//check captive portal option
-	$cp_file='/etc/inc/captiveportal.inc';
-	$pfsense_version=preg_replace("/\s/","",file_get_contents("/etc/version"));
-	$port = ($settings['proxy_port'] ? $settings['proxy_port'] : 3128);
-	$cp_inc = file($cp_file);
-	$new_cp_inc="";
-	$found_rule=0;
-	foreach ($cp_inc as $line) {
-		$new_line=$line;
-		//remove applied squid patch
-		if (preg_match('/skipto 65314 ip/',$line)) {
-			$found_rule++;
-			$new_line ="";
-		}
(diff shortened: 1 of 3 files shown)


doktornotor:
Regarding the second batch of commits trying to undo the b0rkage -- honestly unsure whether it's worth the trouble.

marcelloc:
Doktornoton, this was already fixed when I helped garga to cleanup squid3 package on 2.2.
Without patching captive portal, squid port keeps open for unauthenticated captive portal users.

marcelloc:
And there is a backup file before any changes.

$pfsense_version=preg_replace("/\s/","",file_get_contents("/etc/version"));
if (!file_exists('/root/'.$pfsense_version.'.captiveportal.inc.backup')) {

copy ($cp_file,'/root/'.$pfsense_version.'.captiveportal.inc.backup');

doktornotor:
@marcelloc: "Without patching captive portal, squid port keeps open for unauthenticated captive portal users." 

And WITH patching the file, you completely break the CP it by inserting thousands of empty lines into it.. Please, read the linked forum thread. Sorry, this ain't a viable way to do things.

marcelloc:
until 0.2.8 package version, there bug was an extra "\n" on the patch and it was removed.

I'll do a clean install to teste it again. The forum thread is from april.

marcelloc:
https://github.com/pfsense/pfsense-packages/commit/f9b65f8e27cbc3ae7028d094c70cdc9f38e19523
IMHO it's better fixing it instead of removing it.

doktornotor:
Well yes, of course. But the proper way to fix it is to produce a generic function which allows packages to insert ipfw rules into the CP if they need such thing. There are more proxy packages that can do nontransparent proxy on pfSense than just Squid. So those are broken in the same way, allow CP bypass in the same way and definitely cannot be fixed by another patching of captiveportal.inc, since this gets out of the hands completely even with two packages trying to do this.

PR #909, opened 2015-07-28
Bug #4555 - Tiny DNS: Service does not start

doktornotor:
Attempt on fixing the paths.

### DIFF
--- a/config/tinydns/tinydns.xml
+++ b/config/tinydns/tinydns.xml
@@ -94,62 +94,53 @@
 	<configpath>installedpackages->package->tinydns</configpath>
 	<additional_files_needed>
 		<prefix>/usr/local/pkg/</prefix>
-		<chmod>077</chmod>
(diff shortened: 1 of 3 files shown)


config/tinydns/tinydns.xml
```diff
@@ -2,45 +2,47 @@
...
-    Copyright (C) 2007 to whom it may belong
-    All rights reserved.
-                                                                              */
-/* ========================================================================== */
+	tinydns.xml
+	part of pfSense (https://www.pfSense.org/)
+	Copyright (C) 2009 Bill Marquette
+	Copyright (C) 2010 Ermal Luçi
+	Copyright (C) 2010-2011 Scott Ullrich
+	Copyright (C) 2012-2015 ESF, LLC
```
rbgarga: Where did you get this list of names?

doktornotor: Again, I don't have anything else that the commits history. Considering it was completely missing until now, I figured it's better than leaving it empty. ;)

PR #912, opened 2015-07-29
fix broken textbox and dropdown box

nagyrobi:
icon was missing, and the dropdown had two empty rows when creating new, and one empty row when editing existing value.


nagyrobi:
sorry I see someone else is working on this.

PR #916, opened 2015-07-30
Bug #4731 - softflowd process gets started twice during bootup

doktornotor:
1. Check whether the service is running before trying to restart it. Otherwise it may get started twice on boot by /etc/rc.start_packages.
2. Add standard XML and copyright headers.
3. Remove doubled spaces between sentences in descriptions.


doktornotor:
Actually, this ain't a proper fix. Will create a new one.

PR #917, opened 2015-07-30
softflowd - do not restart on boot plus style fixes

doktornotor:
- Do not try to restart service on boot, otherwise it may get started twice via /etc/rc.start_packages (Fixes bug #4731).
- Add standard XML and copyright headers.
- Remove doubled spaces between sentences in descriptions.
- Developer style guidelines (spacing, braces).

### DIFF
--- a/config/softflowd/softflowd.xml
+++ b/config/softflowd/softflowd.xml
@@ -1,6 +1,45 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE packagegui SYSTEM "./schema/packages.dtd">
+<?xml-stylesheet type="text/xsl" href="./xsl/package.xsl"?>
 <packagegui>
+	<copyright>
+	<![CDATA[
+/* $Id$ */
+/* ========================================================================== */
+/*
+	softflowd.xml
+	part of pfSense (http://www.pfSense.com)
+	Copyright (C) 2013-2015 ESF, LLC
+	All rights reserved.
+*/
+/* ========================================================================== */
+/*
+	Redistribution and use in source and binary forms, with or without
+	modification, are permitted provided that the following conditions are met:
+
+	 1. Redistributions of source code must retain the above copyright notice,
+		this list of conditions and the following disclaimer.
+
+	 2. Redistributions in binary form must reproduce the above copyright
+		notice, this list of conditions and the following disclaimer in the
+		documentation and/or other materials provided with the distribution.
+
+	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+	AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+	AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+	OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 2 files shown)


config/softflowd/softflowd.xml
```diff
@@ -191,39 +243,52 @@
 						"stop" => "/usr/bin/killall -9 softflowd"
 					)
 				);
-				restart_service("softflowd");
+				/* Only try to restart if not booting */
+				if (!($g['booting'])) {
```
rbgarga: Please use platform_booting()

config/softflowd/softflowd.xml
```diff
@@ -145,43 +184,56 @@
 		function sync_package_softflowd() {
 			conf_mount_rw();
 			config_lock();
-			global $config;
+			global $config, $g;
```
rbgarga: This can be undone now you are using platform_booting()

jim-p:
FYI- platform_booting() does not exist on 2.1.x so this PR broke softflowd for anyone trying to use it on 2.1.x. They should upgrade, undoubtedly, but we haven't officially dropped support for packages on 2.1.x yet so we need to try to not break packages there.

doktornotor:
You guys need to sort this mess yourself. Not to mention - some 70+ package changes have been done here without changing those version 8.x versions at all. Probably time to drop the ball.

jim-p:
Once 2.3 is released only then can we officially drop support for 2.1.x packages -- we typically maintain support for current major version + 1 version back. Ideally fixes should be tested on both and the versions bumped on all affected platforms. The PRs really shouldn't have been approved without taking that into consideration, but in most cases it's fine and the impact is minimal.

If a package is known to be broken on 2.1.x and there is no hope of fixing it, then it can be removed from pkg_config.8.* or have its minimum version raised to 2.2.x.

doktornotor:
TBH: this should have been branched for 2.2. Sounds about two years late for that. The damage done by switching to the PBI shit is beyond repair ATM. Plus, some 1/3 of the packages code is dealing with the crap resulting from that and nothing else.

jim-p:
Yep. Can't wait for 2.3 when a lot of this is left in the dust. Hands are sort of tied until then though. But it's coming fast.

doktornotor:
@jim-p This change has been done in quite a few of packages. Frankly, some of them checking for crap like the CSRF token in order to find whether system is booting or not, I cannot see how's it got more broken that that. If someone's going to sync up the package versions in 8.x and 10.x and let people file bugs about those, it can be dealt with. ATM, 2.1,x is de facto abandoned, no matter the official state.

Finally: Branch all these for 2.3. Point 2.3 package manager to that branch. Then people can clean up all the useless code bloat...

rbgarga:
@doktornotor pfsense-packages repo is not going to be used by 2.3, all packages will be part of pfsense/FreeBSD-ports repository.

doktornotor:
@rbgarga That's a good news.Perhaps, it'd be a good idea to move this debate to the forums. Also, most of these 2.1.x/2.2.x conditional hacks could be removed or simplified a lot if a *final* decision is done. For sure would make the code here less buggy and avoid wasting time with nonsense that's not going to be used anyway.

marcelloc:
What about packages that need non default compile options !?

2015-09-10 14:02 GMT-03:00 doktornotor <notifications@github.com>:

rbgarga:
@marcelloc Options can be found at:

https://github.com/pfsense/pfsense/blob/master/tools/conf/pfPorts/make.conf

rbgarga:
@doktornotor for now, to keep things easy to maintain, there is a script that converts a package from pfsense-packages repo into a port. When 2.3 alpha is ready for public tests we will start these discussions.

marcelloc:
great! Thanks garga.

2015-09-10 14:32 GMT-03:00 Renato Botelho <notifications@github.com>:

doktornotor:
@jim-p - All boot checks done "compatible" against in PR #1059 and PR #1060, kindly review and merge.

PR #925, opened 2015-08-01
nut - major code cleanup, code style fixes

doktornotor:
nut.xml and status_nut.php contain just code style fixes per developer style guide.
nut.inc was a **lot** of work

1. nut_notice() and nut_warn() functions removed; use log_error() instead
2. nut_action() function removed; use service_start() and service_stop() instead
3. nut_validate_ip() function removed (WTF); use is_ipaddr() instead
4. replace deprecated eregi() with preg_match()
5. *major* coding style cleanup
6. use full paths to commands in the heredocs

### DIFF
--- a/config/nut/nut.inc
+++ b/config/nut/nut.inc
@@ -2,8 +2,8 @@
 /*
 	nut.inc
 	part of pfSense (https://www.pfsense.org/)
-
-	Copyright (C) 2007 Ryan Wagoner <rswagoner@gmail.com>.
+	Copyright (C) 2007 Ryan Wagoner <rswagoner@gmail.com>
+	Copyright (C) 2015 ESF, LLC
 	All rights reserved.
 
 	Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


config/nut/nut.inc
```diff
@@ -333,7 +326,7 @@ EOD;
 EOD;
 
 		$start = <<<EOD
-if [ `pgrep {$driver} | wc -l` != 0 ]; then
+if [ `/bin/pgrep {$driver} | /usr/bin/wc -l` != 0 ]; then
```
rbgarga: Since you are touching this, you could make this simpler using pgrep return code instead of wc:
```
if /bin/pgrep -q {$driver} 2>/dev/null; then
    {$stop}
fi
```

doktornotor: This single line? Frankly, that pgrep | wc is used consistently in tons of places. I'd rather not touch this code really.

rbgarga: My suggestion didn't use wc, it only uses pgrep return code, that will be 0 (success) if process is running. But nvm, I'll take care of these cases later, maybe change all of them in one shot. I like to change these cases because current code runs multiple commands just to check if process is running. It runs [ (test), then start a new sh session due to backticks, then run pgrep and wc.

PR #934, opened 2015-08-04
pfflowd - code style fixes, add copyright header

doktornotor:
Note: I am NOT bumping the version here, since this is only usable as a disk benchmark ATM. Once Bug #4304 is fixed, it can get a version bump.

P.S. Copyright info taken from git commits history.


cbuechler:
merged, thanks! 

I laughed at the disk benchmark comment. That issue caused about 12.5 MB/sec of disk writes per VM across a couple VMs on one of our dev ESX hosts. Confirmed, it does a fine job of disk stressing. 

I made a separate commit to bump the required version to 3.0 for now to disable it for 10.x base versions since it doesn't do anything useful.

PR #935, opened 2015-08-05
OpenVPN Client Export code style

phil-davis:
Since doctornoktor did a bunch of this, I thought that I might as well review it also. It seems good to make the packages that are "internally" maintained conform to the code style guide.
I bumped the version so it will be easy for people to see exactly what version of code they are running, even though there is no functional change here.


jim-p:
This will need rebased, I made some changes to the package today and the version will need to be bumped again. Also be sure to change the version in pkg_config.8.* as this package is still active on 2.1.x.

rbgarga:
Well, git merged it correctly. Nevermind @phil-davis

PR #942, opened 2015-08-06
Service Watchdog package code style

phil-davis:


config/servicewatchdog/services_servicewatchdog_add.php
```diff
@@ -55,7 +55,7 @@
 
 if ($_POST) {
 	if (!is_numeric($_POST['svcid']))
```
phil-davis: I left this here because I did not want to make any functional change in this pull request.
But this is dodgy-looking - a dangling "if" that I believe will put the following "if" at line 6o effectively inside it.
Looking at the actual code, I don't think it results in any error.
@jim-p - can you look at this and tidy it up or remove however you originally intended it to be?

jim-p: That one is odd. Given what it's testing that should probably have a "return;" on the line after it so that a non-numeric service ID is skipped.

PR #945, opened 2015-08-06
LCDProc Local host fix

topper727:
Seems on my copy of this firewall I have to put the physical ip of local machine as localhost name don't work.Then this package works


topper727:
Merge this please

cbuechler:
Merged, thanks! The reason that ends up failing is lcdproc doesn't bind to IPv6, and localhost will return as ::1 in that case.

topper727:
Thanks for the kudo's.. And merging my fix.. If I knew how to bump the
version I would . I need to watch someone do it.  then I will know

PR #947, opened 2015-08-06
Backup package code style

phil-davis:
I thought I might as well quickly format these little packages also, just so they are done if anyone else ever touches them again. I will look at a couple more like:
blinkled
gwled
LCDproc (maybe)
mailreport
Shellcmd

They seem to be the little packages that are easy and are mostly maintained internally when needed.

### DIFF
--- a/config/backup/backup.inc
+++ b/config/backup/backup.inc
@@ -3,10 +3,10 @@
 /*
 /* ========================================================================== */
 /*
-    backup.inc
-    Copyright (C) 2008 Mark J Crane
-    All rights reserved.
-                                                                              */
+	backup.inc
+	Copyright (C) 2008 Mark J Crane
+	All rights reserved.
+*/
 /* ========================================================================== */
 /*
 	Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


doktornotor:
Well, shellcmd is full of copy-paste disfuntional WTF crap (empty "functions" commented out, some "service" (does not do any service at all) etc. If you want to clean it up, feel free. 

Blinked I did a few days ago, alas apparently the wrong packge. Why do we have two I don't understand. Will redo it now. Sigh.

phil-davis:
I am trying to ignore empty crud... I don't want to break stuff, just quickly format it. If I notice anything that would actually be broken (rather than just NULL) then I care a little bit.

cbuechler:
Always glad to get even style-only fixes, thanks @phil-davis. 

Also thanks @doktornotor for all your package contributions of late. 

We're trying to keep up with the pull requests, but been buried in other work as well this week so a bit slower than I'd like getting things in. 

merged.

phil-davis:
No problem - this sort of crud can sit waiting for a while and it is not stuff that gets touched often, so low chance of merge conflicts.

PR #953, opened 2015-08-07
Fix: Squid RealTime stat 1.20 for the proxy server (unknown) (x.x.x.x:3128).

ReinaldoFeitosa:
Error (1): Cannot get data. Server answered: HTTP/1.1 403 Forbidden

Lines comment is not necessary.

### DIFF
--- a/config/lightsquid/sqstat.php
+++ b/config/lightsquid/sqstat.php
@@ -374,18 +374,18 @@ function sqstat_loadconfig()
     # === load config from pfSense ===
     $iface = '127.0.0.1';
     $iport = 3128;
-    $squid_settings = $config['installedpackages']['squid']['config'][0];
-    if (!empty($squid_settings)) {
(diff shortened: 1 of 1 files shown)


doktornotor:
Uhm... when you get forbidden, perhaps fix the ACLs instead of hardcoding localhost? What if Squid is not running on localhost?

ReinaldoFeitosa:
But IP interface does not solve the Squid is on another host.
What is the purpose of taking the ip of all interfaces? all are equal localhost.

cbuechler:
His point is you don't have to bind squid to 127.0.0.1. Nor should the port be hard coded. There may be a problem of some sort here, but if there is, this isn't a fix. Those lines aren't unnecessary.

ReinaldoFeitosa:
The code just takes the last listed interface on squid.
If you are instaldo on another host will give the same problem.

PR #959, opened 2015-08-08
snort - fix copyright headers, XML cleanup

doktornotor:
Get rid of loads of bad 077 chmods, also no need to chmod 0644 which is default.
Nuke useless, unused XML tags
Fix title and make version consistent.

Done here; not touching anything else here since it's very actively maintained.

@bmeeks8 Hope you don't mind this very small cleanup. ;)

### DIFF
--- a/config/snort/snort.xml
+++ b/config/snort/snort.xml
@@ -3,51 +3,50 @@
 <?xml-stylesheet type="text/xsl" href="../xsl/package.xsl"?>
 <packagegui>
 	<copyright>
-        <![CDATA[
+	<![CDATA[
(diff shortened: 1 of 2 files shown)


bmeeks8:
Thank you for the submissions.  I had wondered about the chmod lines, but since they were legacy and before my time, I had just left them there.

Bill

doktornotor:
Yah, cleaned those from loads of packages... Dunno who started pasting that from where, but it results literally in 0077 permissions, extremely weird plus I don't think "world minus root" writeable files are good for security.

PR #970, opened 2015-08-09
sshdcond.inc - check whether XMLRPC sync is enabled before sync

doktornotor:
```
Starting package SSHDCond...
Warning: Invalid argument supplied for foreach() in /usr/local/pkg/sshdcond.inc on line 127
done.
```

Note: The commit message is wrong, I got confused by having a broken "fix" here locally with a missing closing }


doktornotor:
Damnit, I'm really confused.... The problem basically is:

```
die(var_dump($config['installedpackages']['sshdcondsync']));
NULL
```

This a box which never had this configured. Should even get beyond the first check in the first place. The whole logic is wrong. Closing this, will submit a proper fix.

PR #971, opened 2015-08-09
sshdcond - fix sshdcond_sync_on_changes() logic

doktornotor:
When you install but never configure this package, on boot you get this warning:
```
 Starting package SSHDCond...
Warning: Invalid argument supplied for foreach() in /usr/local/pkg/sshdcond.inc on line 127
done.
```

Problem being that the whole function logic is wrong. Trying to fix here. (Comments included in the code, hopefully the logic is clear and correct now.)

### DIFF
--- a/config/sshdcond/sshdcond.inc
+++ b/config/sshdcond/sshdcond.inc
@@ -117,19 +117,22 @@ function sshdcond_custom_php_write_config() {
 function sshdcond_sync_on_changes() {
 	global $config, $g;
 
-	if (is_array($config['installedpackages']['sshdcondsync'])) {
-		if (!$config['installedpackages']['sshdcondsync']['config'][0]['synconchanges']) {
-			return;
-		}
-	}
-
-	log_error("[sshdcond] xmlrpc sync is starting.");
-	foreach ($config['installedpackages']['sshdcondsync']['config'] as $rs) {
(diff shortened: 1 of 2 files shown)


config/sshdcond/sshdcond.inc
```diff
@@ -117,23 +117,26 @@ function sshdcond_custom_php_write_config() {
...
-	foreach ($config['installedpackages']['sshdcondsync']['config'] as $rs) {
-		foreach($rs['row'] as $sh) {
-			$sync_to_ip = $sh['ipaddress'];
-			$password = $sh['password'];
-			if ($password && $sync_to_ip) {
-				sshdcond_do_xmlrpc_sync($sync_to_ip, $password);
+	if (!is_array($config['installedpackages']['sshdcondsync'])) {
+		/* Basically, this package was never configured */
+		return;
+	} elseif (!$config['installedpackages']['sshdcondsync']['config'][0]['synconchanges']) {
```
rbgarga: Actually you don't need elseif here, because if it entered in above if, it's going to return. You also don't need that else on line 126. You will end up with something like:
```
if (!is_array(...)) {
    return;
}
if (!isset(...)) {
    return;
}
/* Do XMLRPC sync */
```

doktornotor:
OK, lets try again. :)

PR #972, opened 2015-08-10
add "dev" option for inline config

ledowong:
so it works on both iOS and Tunnelblick

### DIFF
--- a/config/openvpn-client-export/openvpn-client-export.inc
+++ b/config/openvpn-client-export/openvpn-client-export.inc
@@ -207,10 +207,8 @@ function openvpn_client_export_config($srvid, $usrid, $crtid, $useaddr, $verifys
 
 	// add basic settings
 	$devmode = empty($settings['dev_mode']) ? "tun" : $settings['dev_mode'];
-	if (($expformat != "inlinedroid") && ($expformat != "inlineios")) {
-		$conf .= "dev {$devmode}{$nl}";
-	}
-	if (!empty($settings['tunnel_networkv6']) && ($expformat != "inlinedroid") && ($expformat != "inlineios")) {
+	$conf .= "dev {$devmode}{$nl}";
+	if (!empty($settings['tunnel_networkv6'])) {
 		$conf .= "tun-ipv6{$nl}";
 	}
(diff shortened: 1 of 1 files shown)


doktornotor:
What problem are you solving here? It already works when you export it in the *proper* format.

ledowong:
Which format is *proper* for Tunnelblick?

I guess 'dev' option is common in inline config file.
I tried Synology build-in OpenVPN, [OpenVPN Virtual Appliances](https://openvpn.net/index.php/access-server/download-openvpn-as-vm.html) and ASUS router build-in OpenVPN, all their config file with 'dev' option.

I see no harm to include 'dev' option back to inline config file. It doesn't break iOS/Andriod. It just make the config file compatible with more client. Btw it was [there](https://github.com/pfsense/pfsense-packages/commit/34664d0519236cd45ab223963edb2e87aa175804).

doktornotor:
That's what not what you have patched there. You have changed not only the dev- option. Now, you also dump tun-ipv6 in all  inline configurations.

jim-p:
The original change you commented on was two and a half years old. We've used the clients in question since then without problems. Perhaps there is something else wrong in your client/config, or you used the wrong format. Either way, as has been pointed out, the change in this pull request does much more than stated. This needs some discussion on the forum and much more agreement before being pulled in.

ledowong:
I afraid I am using the correct client/config (inline config for Tunnelblick), following [this](https://doc.pfsense.org/index.php/OpenVPN_Client_Export_Package) document.

>**Inline Configurations**
>* Others: Exports an inline configuration file suitable for use by any other current OpenVPN client, such as the Windows client, Viscosity, or Tunnelblick.
>
>**Mac OSX**
>The current versions of Viscosity and Tunnelblick also import Inline configurations properly.

But it doesn't work. It might be [Tunnelblick](https://tunnelblick.net/downloads.html) make some changes in this two and a half years, I am not sure, I am using version 3.5.3.
Base on the log from Tunnelblick, it said 'dev' option is needed. So I edit the config file, add 'dev' option, and it works.

Yea @doktornotor is right, I shouldn't touch 'tun-ipv6', or maybe android as well. Thanks.

PR #977, opened 2015-08-10
Remove useless openvpn-tapfix "package"

doktornotor:


jim-p:
Technically it's still valid for someone on 2.0.x, but at this point nobody should be running that, so it's OK to ditch this. If someone really needs it they can install the patch by hand.

doktornotor:
Yeah, installing it by system patches or such, it might even be "uninstalled", unlike this one. :)

PR #978, opened 2015-08-10
Remove useless pure-ftpd package

doktornotor:
maximum_version 1.0 - wow...


cbuechler:
that was overdue indeed, appreciate all the clean up!

PR #979, opened 2015-08-10
snort, fix config-sync to 'system backup server'

PiBa-NL:
snort, fix config-sync to 'system backup server'

### DIFF
--- a/config/snort/snort.inc
+++ b/config/snort/snort.inc
@@ -3761,8 +3761,8 @@ function snort_sync_on_changes() {
 				}
 				break;
 			case "auto":
-					if (is_array($config['installedpackages']['carpsettings']) && is_array($config['installedpackages']['carpsettings']['config'])){
-						$system_carp=$config['installedpackages']['carpsettings']['config'][0];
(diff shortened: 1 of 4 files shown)


rbgarga:
@bmeeks8 can you please take a look at this patch?

bmeeks8:
I see no problem with the patch except now with the recent merge of my "3.2.7" update the version number for this patch would need to change to 3.2.8.  Alternatively, it is probably OK to leave the version the same (at 3.2.7) and folks impacted who wanted the update could simply do a GUI package reinstall.

doktornotor:
When you are already touching this, mind fixing the instances of ```$snort_sync[row]``` -> ```$snort_sync['row']``` as well a couple of lines above? (One of the endless copies of marcelloc's XMLRPC sync code in the repo, all of which are slowly diverting from each other and have different level of bugs/lack of validation. Sigh... https://forum.pfsense.org/index.php?topic=61586.msg548518#msg548518)

rbgarga:
Done @doktornotor

PR #986, opened 2015-08-11
Disable strikeback package on 2.2

doktornotor:
This package is completely broken, calling pkg_add/pkg_delete, generally WTF code.

### DIFF
--- a/pkg_config.8.xml
+++ b/pkg_config.8.xml
@@ -89,6 +89,8 @@
 		<version>0.1</version>
 		<status>BETA</status>
 		<required_version>2.0</required_version>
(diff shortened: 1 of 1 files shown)


jim-p:
FYI- 2.2 wouldn't read pkg_config.8.xml*, only pkg_config.10.xml, so this didn't have any net effect.

rbgarga:
Yeah, you are right @jim-p, I misread the file name. At least it won't break anything :)

doktornotor:
Errr... yeah, that's not what I meant to commit though. I'll resubmit this, this commit was a waste of time indeed. :smile:

PR #987, opened 2015-08-11
Disable strikeback package on 2.2 (for real now)

doktornotor:
This package is completely broken, calling pkg_add/pkg_delete, generally WTF code.

### DIFF
--- a/pkg_config.10.xml
+++ b/pkg_config.10.xml
@@ -96,7 +96,9 @@
 		<category>Services</category>
 		<version>0.1</version>
 		<status>BETA</status>
-		<required_version>2.2</required_version>
+		<required_version>2.1</required_version>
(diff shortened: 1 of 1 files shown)


jim-p:
Even better -- just remove it from this file entirely, since 2.1.x can't/won't read this file. :-)

doktornotor:
OK, guys. Someone please hide the *beep* from people in whatever manner you find appropriate. I meant to keep it there in case someone eventually might get mad enough to fix the thing (definitely NOT myself), and it felt wrong to have required version higher than max. version. So, it was not to suggest that 2.1.x should read this file.

jim-p:
OK, fair enough. Usually in those cases we set the version requirements absurdly low or absurdly high and note in the comment to "fix/remove once the package has been unbroken" or something to that effect.

doktornotor:
Ok, looks like done with cc096698 so closing this. Thanks again. :+1:

PR #998, opened 2015-08-12
vhosts - rewrite garbage package

doktornotor:
Seriously... WTF!!! (Completely untested beyond PHP syntax parser, since the package is so broken there's absolutely no chance it'd work for anyone.) Fetching and unpacking tarballs with PHP, just wow.


cbuechler:
merged, thanks!

PR #1016, opened 2015-08-20
Altered OpenVPN Client Export

KyleJohnstonNet:
I altered OpenVPN Client Export to include entire CA chain when exporting, rather than just the server certificate's immediate CA. This corrects issues were exported configurations fail due to the server's cert being issued by an intermediate CA. (See: https://redmine.pfsense.org/issues/4934)

This change works as expected for my use case (where the server's certificate is the child of an intermediate certificate) but has not been tested when the server's certificate is the direct child of a CA, or when the server's certificate is itself a CA.

### DIFF
--- a/config/openvpn-client-export/openvpn-client-export.inc
+++ b/config/openvpn-client-export/openvpn-client-export.inc
@@ -138,7 +138,15 @@ function openvpn_client_export_validate_config($srvid, $usrid, $crtid) {
 	if (!$server_cert) {
 		$input_errors[] = "Could not locate server certificate.";
 	} else {
+<<<<<<< Updated upstream
 		$server_ca = lookup_ca($server_cert['caref']);
+=======
+		/*
+			Replaced lookup_ca($server_cert['caref']) with ca_chain($server_cert). Note that ca_chain returns a string of Base64 encoded certificates, not an array. 
+			Thus, all references to base64_decode($server_ca['crt']) have been replaced with simply $server_ca
+		*/
+		$server_ca = ca_chain($server_cert); 
+>>>>>>> Stashed changes
 		if (!$server_ca) {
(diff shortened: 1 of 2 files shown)


config/openvpn-client-export/openvpn-client-export.inc
```diff
@@ -138,7 +138,15 @@ function openvpn_client_export_validate_config($srvid, $usrid, $crtid) {
...
 	} else {
+<<<<<<< Updated upstream
 		$server_ca = lookup_ca($server_cert['caref']);
+=======
+		/*
+			Replaced lookup_ca($server_cert['caref']) with ca_chain($server_cert). Note that ca_chain returns a string of Base64 encoded certificates, not an array. 
+			Thus, all references to base64_decode($server_ca['crt']) have been replaced with simply $server_ca
+		*/
+		$server_ca = ca_chain($server_cert); 
+>>>>>>> Stashed changes
```
rbgarga: I'm not sure what you are trying to submit, but looks like you didn't deal right with a merge conflict.

KyleJohnstonNet: This was incorrect in this commit. I tried to remove it from the chain, but failed.

The code is correct in the final commit: 06ec0438662c73a2d70b615a9c40f5c0d30b3bd1.

rbgarga: If this is the only commit, you can save it externally, then start a new branch, apply it, and re-submit the Pull Request

rbgarga: And btw, the commit you mentioned just added comments but didn't replace lookup_ca() by ca_chain() as mentioned in the comment. Can you please send a new Pull Request with the correct patch?

rbgarga:
Re-submitted as #1029

PR #1024, opened 2015-08-25
Remove useless remount calls

doktornotor:
This removes loads of pointless conf_mount_ro() and conf_mount_rw() calls from custom install/deinstall functions in packages since install_package() and uninstall_package() from pkg-utils.inc already takes care of this. Should really help to reduce the install failures and slowness mess especially on nanobsd with recent pfSense versions.

Also, bunch of other redundant stuff is removed in a couple of packages, like, stopping defined service on uninstall and removing its rcfile, those are already handled by delete_package_xml() from pkg-utils.inc.


phil-davis:
I will just make the observation that conf_mount_rw/conf_mount_ro keep a count of the number of "times/reasons" that the file system is mounted RW. After the first conf_mount_rw() call, the file system is mounted RW and the count is 1. The next conf_mount_rw() call only increments the counter in shared memory to 2 - it is almost a no-op. Then the next conf_mount_ro() decrements the counter back to 1 - also almost a no-op. Only when the counter goes back to zero is the actual file system mount RO done, which has the time delay.
So these conf_mount_rw/conf_mount_ro pairs should not be causing any significant time delay.
If the function they are in is only called from the package installer (or other places that already have done conf_mount_rw) then fine, they are not needed in real life. But they do "protect" from a caller that might call the function without having first done conf_mount_rw().

doktornotor:
None of these functions get called anywhere else from the packages, this is install/uninstall only. (Also, did not touch general mess such as Squid* packages family :angry: at all, since I don't have time to dig into the junk.)

These are trully pointless and getting copied over to other packages since people think it's a good example how to do things.

PR #1030, opened 2015-08-29
avahi - expose many more config options in GUI

doktornotor:
Also, switched "Disable IPv4/IPv6" to "Enable IPv4/IPv6" (both enabled by default). Added basic input validation.

### DIFF
--- a/config/avahi/avahi.inc
+++ b/config/avahi/avahi.inc
@@ -223,7 +223,7 @@ function avahi_sync() {
 	}
 	avahi_write_config();
 	// Is package enabled?
-	if (($config['installedpackages']['avahi']['config'][0]['enable']) && file_exists("/usr/local/etc/rc.d/avahi-daemon.sh")) {
+	if ($config['installedpackages']['avahi']['config'][0]['enable']) {
 		start_service("avahi");
 	}
 }
--- a/config/avahi/avahi.inc
+++ b/config/avahi/avahi.inc
@@ -233,7 +233,7 @@ function avahi_sync() {
 function avahi_upgrade_config() {
 	global $config, $avahi_config;
 
-	$avahi_config &= $config['installedpackages']['avahi']['config'][0];
+	$avahi_config =& $config['installedpackages']['avahi']['config'][0];
 	if (!is_array($avahi_config)) {
 		$avahi_config = array();
 	}
--- a/config/avahi/avahi.inc
+++ b/config/avahi/avahi.inc
@@ -233,27 +233,25 @@ function avahi_sync() {
 function avahi_upgrade_config() {
 	global $config, $avahi_config;
 
-	$avahi_config = $config['installedpackages']['avahi']['config'][0];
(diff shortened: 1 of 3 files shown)


config/avahi/avahi.inc
```diff
@@ -63,23 +63,34 @@ function avahi_deinstall() {
 }
 
 function avahi_write_config() {
-	global $config;
+	global $config, $avahi_config;
 	conf_mount_rw();
 
 	// Pull some various values out of config.xml
+	$avahi_config = $config['installedpackages']['avahi']['config'][0];
```
rbgarga: It would be good to check if this $config section isset() before assign it.

config/avahi/avahi.inc
```diff
@@ -203,9 +223,49 @@ function avahi_sync() {
...
+function avahi_upgrade_config() {
+	global $config, $avahi_config;
+
+	$avahi_config = $config['installedpackages']['avahi']['config'][0];
+	if (!is_array($avahi_config)) {
+		$avahi_config = array();
+	}
+	$changes = 0;
+	// Convert previous disable values to enable ones
+	if ($avahi_config['disable_ipv4'] != "on" ) {
```
rbgarga: Same here, check if key isset() before attempt to change it. Also, you probably want to assign $avahi_config =& to be able to change $config values, and unset old keys

config/avahi/avahi.inc
```diff
@@ -203,9 +225,49 @@ function avahi_sync() {
...
+	if (isset($avahi_config['disable_ipv4']) && $avahi_config['disable_ipv4'] != "on") {
+		$avahi_config['enable_ipv4'] == "on";
+		$changes++;
+	}
+	if (isset($avahi_config['disable_ipv6']) && $avahi_config['disable_ipv6'] != "on") {
+		$avahi_config['enable_ipv6'] == "on";
+		$changes++;
+	}
+	// If broken settings were configured due to lack of validation, just set to defaults
+	if (isset($avahi_config['disable_ipv4']) && isset($avahi_config['disable_ipv6']) && $avahi_config['disable_ipv4'] == "on" && $avahi_config['disable_ipv6'] == "on") {
```
rbgarga: It's still missing unset() calls for old config knobs, otherwise it's going to keep it in config unnecessarily

config/avahi/avahi.inc
```diff
@@ -203,9 +225,53 @@ function avahi_sync() {
...
-	if (($config['installedpackages']['avahi']['config'][0]['enable']) && file_exists("/usr/local/etc/rc.d/avahi-daemon.sh")) {
+	if ($config['installedpackages']['avahi']['config'][0]['enable']) {
 		start_service("avahi");
 	}
 }
 
+function avahi_upgrade_config() {
+	global $config, $avahi_config;
+
+	$avahi_config = $config['installedpackages']['avahi']['config'][0];
```
rbgarga: you should use reference (&=) here, otherwise unset() you call below won't have any effect on $config

PR #1031, opened 2015-08-29
widget-antivirus - fix havp version check, nuke useless junk

doktornotor:


config/widget-antivirus/antivirus_status.widget.php
```diff
@@ -123,8 +104,16 @@ function dwg_av_statistic() {
...
-				echo exec("pkg_info | grep \"[h]avp\"");
+				// HAVP version
+				$pfs_version = substr(trim(file_get_contents("/etc/version")), 0, 3);
+				if ($pfs_version == "2.1") {
+					echo exec("pkg_info | grep \"[h]avp\"");
+				} elseif ($pfs_version == "2.2") {
+					// Show package version at least, no good quick way to get the PBI version
+					echo "pkg v{$config['installedpackages']['package'][get_pkg_id("havp")]['version']}";
+				} else {
+					echo exec("/usr/sbin/pkg info havp | /usr/bin/head -n 1");
```
rbgarga: You can use pkg_exec() here to run 'pkg query "%v" havp' that will return only version

doktornotor: Just to be sure, this is referring to 2.3+? Kinda confused with the packaging transitions already.

rbgarga: Exactly

doktornotor: OK, hopefully done ;)

rbgarga:
pkg_exec() prototype is:
```
int pkg_exec($params, &$stdout, &$stderr)
```
it returns pkg call return code. Version you are expecting will be present in stdout variable.

doktornotor:
Well, last attempt. Obviously cannot test this anywhere, plus not really sure what am I supposed to to with &$stderr which I don't need at all.

PR #1038, opened 2015-09-04
spamd - code style cleanup and multiple fixes

doktornotor:
spamd.inc
- Code style and indentation fixes
- Use install_cron_job() instead of re-implementing it
- Use restart_service()
- Clean up user/groups after itself on uninstall (fixes Bug #4929)
- Fix input validation (no errors shown due to a typo in variable name), and require IPv4 for NextMTA. No aliases allowed since it was producing broken firewall rules and is generally pointless for a single IP.
- Better log messages plus some cosmetics

spamd.xml
- Update copyright header
- Fix file permissions (no need for these to be executable)
- Fix whitespace/indentation
- Nuke useless comments
- Sanitize package version

spamd_db.php
- Code style and indentation fixes
- Use shell_exec() instead of backticks
- Don't add the same email address to spamtrap twice (?!)
- XHTML fixes

spamd_settings.xml
- Fix copyright header
- Remove service and menu tags that are already specified in packages configfile (spamd.xml)
- Nuke useless comments
- Make descriptions readable
- Fix default values handling
- Remove completely bogus "Enable RRD graphing" option that had no code attached to it.
- Indentation/whitespace fixes
- Require IPv4 for NextMTA

spamd_whitelist.xml
- Fix copyright header
- Remove useless comments
- Indentation/whitespace fixes

spamd_db_ext.php, spamd_outlook.xml, spamd_gather_stats.php, spamd_exchexp.asp, spamd_verify_to_address.php, spamd_rules.php
- Deleted (dead unused code)


config/spamd/spamd_db.php
```diff
@@ -122,35 +119,38 @@
 }
 
 /* spam trap e-mail address */
-if($_GET['spamtrapemail'] <> "") {
```
doktornotor: @PiBa-NL You've been fixing "add spamtrap" in 7aac4188282948ac508a1fa3b625dbf514e9d0e4  - I don't get this. Why was it being done twice?

PiBa-NL: Good spotting, to be honest i don't know. As you can see it was already done like that also before i touched the code. But i think your right, it should only be done once.. At least after the edits it worked better than before.. Still fundamentally several things wrong with this package though, but that's a different subject.

doktornotor: Well, I just wanted to ask since perhaps it might have been added on purpose. Doesn't seem so. Thanks! ;)

PR #1044, opened 2015-09-05
Update vpn_openvpn_export.php

gavit:
Windows installers reflect the target OS it's for.
e.g.: x86-win6, x86-xp, x64-win6, x64-xp


jim-p:
As-is, this is not complete enough and would be a potential security problem. Before putting $act into a filename it must be sanitized/validated to ensure it only contains proper characters and that it does not contain any special characters such as path separators/specifications, etc, etc. Even better would be to validate that it is exactly equal to one of the options specified in the GUI.

PR #1047, opened 2015-09-06
squid - use install_cron_job() instead of reimplementing it

doktornotor:


doktornotor:
Accidentally posted Squid3 file here, will resubmit.

PR #1052, opened 2015-09-08
Patch 1

ghost:
Sanitized code to protect against potential security issues.

### DIFF
--- a/config/openvpn-client-export/vpn_openvpn_export.php
+++ b/config/openvpn-client-export/vpn_openvpn_export.php
@@ -252,20 +252,20 @@ if (!empty($act)) {
 		$exp_path = viscosity_openvpn_client_config_exporter($srvid, $usrid, $crtid, $useaddr, $verifyservercn, $randomlocalport, $usetoken, $password, $proxy, $openvpnmanager, $advancedoptions);
 	}
 
-	if (substr($act, 0, 11) == "inst-x86-xp") {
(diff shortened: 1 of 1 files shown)


jim-p:
It's better, but not ideal. Lots of unnecessarily duplicated code. With some better structure that could be reduced significantly. Also there would be no need to have "inst-" and "install" in the filename. If you're hardcoding it, may as well reduce the redundancy.

rbgarga:

gavit:
Please see attached,

Kind regards,
Gabor

config/openvpn-client-export/vpn_openvpn_export.php
```diff
@@ -252,8 +252,20 @@
 		$exp_path = viscosity_openvpn_client_config_exporter($srvid, $usrid, $crtid, $useaddr, $verifyservercn, $randomlocalport, $usetoken, $password, $proxy, $openvpnmanager, $advancedoptions);
 	}
 
-	if (substr($act, 0, 4) == "inst") {
-		$exp_name = urlencode($exp_name."-install.exe");
+	if (substr($act, 0, 6) == "inst-x") {
+		if (substr($act, 6, 9) == "86-") {
+			if (substr($act, 9, 11) == "xp) {
```
phil-davis: Missing closing double-quotes.
Same syntax error in 4 places.

rbgarga:

PR #1065, opened 2015-09-14
Remove Snort package for 2.1.x

doktornotor:
The compatible Snort VRT rules version can no longer be downloaded (and the ruleset version is hardcoded for each Snort version) making the package completely useless. https://www.snort.org/eol

### DIFF
--- a/pkg_config.8.xml
+++ b/pkg_config.8.xml
@@ -459,26 +459,6 @@
 		<configurationfile>vhosts.xml</configurationfile>
 	</package>
 	<package>
-		<name>snort</name>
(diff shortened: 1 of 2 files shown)


rbgarga:
Any thoughts about it @bmeeks8 ?

bmeeks8:
I agree.  The Snort VRT has discontinued rules package support for the Snort version currently in the pfSense 2.1.x package repository.  This means users cannot download any rules from the Snort VRT web site (the file for the 2.9.7.2 Snort version has been pulled).

cbuechler:
Merged, thanks!

PR #1070, opened 2015-09-16
ntopng - fix up 2.1.x vs. 2.2.x PBI mess, make $ntopng_config global in ntopng_sync_package()

doktornotor:
Put $ntopng_config into globals, might fix the service restart after package reinstall...

Note: If someone's going to rebuild the PBIs for https://redmine.pfsense.org/issues/4831 in a reasonable timeframe, this one can definitely wait for that. :wink:

### DIFF
--- a/config/ntopng/ntopng.inc
+++ b/config/ntopng/ntopng.inc
@@ -151,8 +151,10 @@ function ntopng_sync_package() {
 
 	$start = "";
 	$stop = "";
-	if ($pf_version == "2.1" || $pf_version == "2.2") {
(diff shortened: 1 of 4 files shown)


doktornotor:
Just found out that the paths got somehow messed up long before splitting the PHP code to separate inc file, breaking the GeoIP updates, among others. Hopefully fixed all with the latest commit. Someone kindly double-check with fresh eyes, I'm seriously tired of the PBI junk. For 2.2.x, all the PBI paths should be prefixes with extra ```/local```, on 2.1.x there shouldn't be any such thing.

PR #1077, opened 2015-09-19
squidguard/squidguard-devel - do not breaks blacklists with categories as symlink (Bug #3085)

doktornotor:


doktornotor:
Eh... problem is one level above.

PR #1081, opened 2015-09-26
Clarify Squid 2.7 status

doktornotor:
This thing is badly broken, dead as a coffin-nail, hasn't been touched for years, noone will fix it and noone should be using it.

### DIFF
--- a/pkg_config.10.xml
+++ b/pkg_config.10.xml
@@ -1022,11 +1022,15 @@
 	</package>
 	<package>
 		<name>squid</name>
-		<descr>High performance web proxy cache.</descr>
+		<descr><![CDATA[
+			High performance web proxy cache.<br/>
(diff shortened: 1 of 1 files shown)


jim-p:
The package itself still works fine, it's one we officially claim to support. Customers still use it regularly and in many cases it works better than the squid 3.x package (though mostly because you can't break it in ways that you can break squid 3.x) -- once we get to 2.3 where there is no more squid 2.7 and the squid 3 package is good, then this can be done, but not yet.

doktornotor:
The package certainly is wonderful, except that you cannot even save settings. https://redmine.pfsense.org/issues/5205

If you officially claim to support this broken thing, then you'll need to fix it.

jim-p:
The problem saving there was actually something that shouldn't have been working on 2.2.x either. Just luck that it wasn't throwing the same error the way the variables were being referenced. I pushed a fix.

PR #1082, opened 2015-09-27
Add privileges configuration to packages - part 1

doktornotor:
This adds configurable privileges for packages to System - User Manager.


pfSense-git-updates:

doktornotor:
Huh? How many times should I sign it?

rbgarga:
@doktornotor please ignore the message, we are testing a script to automatically add these messages and set labels, but it still needs some fixes

doktornotor:
Ah, OK... :)

PR #1087, opened 2015-10-02
Quagga webgui fix for 2.3-alpha

heper:
see https://forum.pfsense.org/index.php?topic=100249.0 
https://redmine.pfsense.org/issues/5239

it appears you need a "listtopic" field with a name or else addInput spawns an error. (perhaps it can be worked around in another way aswell)

if this is not the intended behaviour, then it might be better if someone took a look at addInput() and work around it there.


heper:
in favour of https://github.com/pfsense/pfsense/commit/86a20575ca00ca6ee7c36977289b1bbe66dea3b2

PR #1094, opened 2015-10-12
haproxy1_5: remove weird visibility:hidden when a listener is secondary

thib-ack:
With the current version of haproxy1_5 package, I have this behavior in the interface when having shared frontends (in the frontends section) :
current

With the patch I get something more normal :
withpatch


pfSense-git-updates:

PR #1096, opened 2015-10-14
haproxy1_5: add the port option for health-checks in pools interface

thib-ack:
In HAProxy1_5 package, handle the option "port" on backends health-checks to use a custom port. For example HAproxy is load-balancing on my webservers on port 80 but I have a custom check service available on port 9200.

See documentation :
http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#port

check-port-option


pfSense-git-updates:

PR #1098, opened 2015-10-15
Add privileges configuration to squidguard package

doktornotor:
Hey guys, kindly check that whatever thing that parses this swallows multiple &lt;conflicts&gt; tags per package. Couldn't find any code anywhere regarding this - at all.

### DIFF
--- a/config/squidGuard/squidguard.priv.inc
+++ b/config/squidGuard/squidguard.priv.inc
@@ -1,0 +1,55 @@
+<?php
+/*
+	squidguard.priv.inc
+	part of pfSense (http://www.pfSense.org/)
+	Copyright (C) 2015 ESF, LLC
+	All rights reserved.
+
+	Redistribution and use in source and binary forms, with or without
+	modification, are permitted provided that the following conditions are met:
+
+	1. Redistributions of source code must retain the above copyright notice,
+	   this list of conditions and the following disclaimer.
+
+	2. Redistributions in binary form must reproduce the above copyright
+	   notice, this list of conditions and the following disclaimer in the
+	   documentation and/or other materials provided with the distribution.
+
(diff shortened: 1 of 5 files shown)


doktornotor:
If this cannot be done, I'm frankly inclined to stick some hacks into custom_php_install_command() to cause the package install to fail if conflicting packages are found. These things mess up each other, and mess up non-matching Squid versions badly.

rbgarga:
Please use a single conflicts tag with all items separated by space. I'll take care of introducing necessary checks. Thanks!

doktornotor:
OK, I'll redo this shortly, thanks :)

doktornotor:
Hmmm... "separated by space" is rather unfortunate when the name tags have spaces in package names. WTF. Should this be quoted (single, double) or what? Sigh.

rbgarga:
Yeah, you are right @doktornotor. Let it as is, with multiple conflicts tags, and I'll take care and change the script that creates 2.3 ports

doktornotor:
Well, another alternative would be renaming the packages (this was a very bad idea in the first place to allow spaces there) but definitely too late for that.

rbgarga:
I'm working to merge this one, but I reverted conflicts changes for now. This tag conflicts is only used by the script that convert a 2.2 package into a 2.3 port on repo FreeBSD-ports, and since we are not converting all those packages versions to 2.3, we don't need to do it.

The only method 2.2 has to control conflicts is based on tag internal_name, and this is not enough unfortunately to cover all those cases you figured out. It's too late to make a big change on 2.2, lets concentrate to have a better package system on 2.3

Thanks once again for all your contributions!!!

doktornotor:
OK... I guess I'll do a new PR to add some warnings into package descriptions instead, I guess that'd be OK and cause no compatibility issues, also doesn't require any changes elsewhere.

Thanks. :)

PR #1103, opened 2015-10-18
Avahi - Logic for publishing was swapped.  disable checkbox was enabling & vice-versa

Morlok8k:
Basically if I wanted to publish services via Avahi, I have to check "Disable publishing" and "Disable User Service Publishing", because they operate backwards compared to the description in the web-gui.

This minor fix should change it so it operates according to the description.

### DIFF
--- a/config/avahi/avahi.inc
+++ b/config/avahi/avahi.inc
@@ -82,8 +82,8 @@ function avahi_write_config() {
 	// Wide Area
 	$widearea = ($avahi_config['enable_wide_area']) ? "yes" : "no";
 	// Publishing Options
-	$publish = ($avahi_config['disable_publishing']) ? "no" : "yes";
-	$userpublish = ($avahi_config['disable_user_service_publishing']) ? "no" : "yes";
+	$publish = ($avahi_config['disable_publishing']) ? "yes" : "no";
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

Morlok8k:
I have signed an CLA electronically.  Should be good now.

doktornotor:
Pardon me, but this commit is completely absurd -- how on earth was this operating backwards? So, you have tons of enable checkboxes in the GUI, which now use the same logic as the "disable" ones?

Morlok8k:
doktornotor deleted his comment but it said:
> Pardon me, but this commit is completely absurd -- how on earth was this operating backwards? So, you have tons of enable checkboxes in the GUI, which now use the same logic as the "disable" ones?

Basically the end result was swapped:
with those two checked we got:
> [publish]
> disable-publishing=NO
> disable-user-service-publishing=NO

and we want that result with them unchecked.

The tricky part here is the disable dbus checkbox.  it is backwards too, but functions correctly.  all the text on the GUI says disable, the logic is swapped compared to the others, but the real configuration file says "enable-dbus".  So, it all works correctly.

I assume since he deleted his comment, he realized his mistake.

PR #1104, opened 2015-10-18
lightsquid package reworked

doktornotor:
- provide instructions to users and add corresponding input validation; completely stop touching Squid packages configuration directly
- fix cronjob handling
- fix 'Skip URL(s)' handling and add some basic validation to it as well
- fix lightsquid.cfg handling
- fix English and clean up the GUI
- fix file permissions
- add privileges configuration
- code style fixes
- HTML fixes
- ...


doktornotor:
As noted above, what the heck is that assignment doing inside the if() ?!

PR #1111, opened 2015-10-24
haproxy-devel, pkg v0.32 allow usage of mailers/dns/'actions', rework of acl system.

PiBa-NL:
haproxy-devel, pkg v0.32 allow usage of mailers/dns/'actions', rework of acl system.

### DIFF
--- a/config/haproxy-devel/haproxy.priv.inc
+++ b/config/haproxy-devel/haproxy.priv.inc
@@ -1,0 +1,50 @@
+<?php
+/*
+	haproxy.priv.inc
+	part of pfSense (http://www.pfSense.org/)
+	Copyright (C) 2015 ESF, LLC
+	All rights reserved.
+
+	Redistribution and use in source and binary forms, with or without
+	modification, are permitted provided that the following conditions are met:
+
+	1. Redistributions of source code must retain the above copyright notice,
+	   this list of conditions and the following disclaimer.
+
+	2. Redistributions in binary form must reproduce the above copyright
+	   notice, this list of conditions and the following disclaimer in the
+	   documentation and/or other materials provided with the distribution.
+
+	THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+	INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+	AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+	AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+	OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+	SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+	INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+	CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+	ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+	POSSIBILITY OF SUCH DAMAGE.
+*/
+global $priv_list;
+
+$priv_list['page-service-haproxy'] = array();
+$priv_list['page-service-haproxy']['name'] = "WebCfg - Services: HAProxy package";
+$priv_list['page-service-haproxy']['descr'] = "Allow access to HAProxy package GUI";
+$priv_list['page-service-haproxy']['match'] = array();
+
+$priv_list['page-service-haproxy']['match'][] = "haproxy_files.php*";
+$priv_list['page-service-haproxy']['match'][] = "haproxy_global.php*";
+$priv_list['page-service-haproxy']['match'][] = "haproxy_listeners.php*";
+$priv_list['page-service-haproxy']['match'][] = "haproxy_listeners_edit.php*";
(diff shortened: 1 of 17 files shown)


justnx:
latest haproxy-devel pkg v0.32 is crashing with message: pid 32030 (haproxy), uid 0: exited on signal 11
Any quick fix for that?

PiBa-NL:
new pull-request is pending which should be using a different haproxy version..
im no really easy fix though..
you might want to pbi_delete the current package (pbi_info will show you which is installed). then use fetch do download and pbi_add to install http://files.pfsense.org/packages/10/All/haproxy-devel-1.6-dev2-amd64.pbi

Sorry, kinda falling in the middle of versions i think..

PiBa-NL:
Even so 1.6-dev4 (on settings tab it shows the version)  should work to.. at least a single backend works with a few test requests for me..

PiBa-NL:
@justnx , actually when exactly is it crashing? during startup? or after a number of requests? could you post some info on redmine or forum with your configuration?

PiBa-NL:
b.t.w. there is of course the easy fix of using haproxy-1_5 package.. assuming that you have a config backup and don't need and 1.6 features..

justnx:
i went back to haproxy-1_5 since 1.6 keeps crashing short after some requests.
Btw. is there any possibility to implemente the latest stable branch: 1.6.2-1 in replacement for the haproxy-1_5 stable?

PiBa-NL:
Yes 1.6.2 stable pullrequest is pending.. should be available in a day or two (i hope) If it works and no issues are found ill likely create a haproxy-1_6 package..

PiBa-NL:
@justnx if your still in testing phase or have a window to try, could you checkout out the haproxy-devel package v0.33 ? It uses 1.6.2_1 release version of haproxy. Instead of what was still dev4 a few days ago..

justnx:
sure i can test.

Before i can extend test it, i need to figure out how i can migrate my old 1.6 config to new format.
Any idea? Btw. where is the loaded haproxy.cfg stored?

Btw. i realized two common bugs in the new GUI
The Mailserver and DNS-Server field is choosen as default but without a value in it, which causing syntax error on saving the config.

PiBa-NL:
/var/etc/haproxy/haproxy.cfg or while checking /var/etc/haproxy-test/haproxy.cfg
The empty dns/mail server records should not be added anymore when upgrading..

justnx:
Okay, i tested it so far. The GUI tabs e.g. and migration seems fine now. All is on its place.

But it seems that there is something wrong in the tcp/kernel stack with this HAProxy binary.
Mysql Master/Failover over TCP balancing setup seems to not doing its job well.

I'm getting alot PHP Error on very small Server load:
Nov  7 19:10:30 ngx-03 err: 2015/11/07 19:10:30 [error] 7327#7327: *225 FastCGI sent in stderr: "PHP message: PHP Warning:  Packets out of order. Expected 1 received 0. Packet size=0 in /storage/www/xxxxxx/xxx_MySQL.class.php on line 185#012PHP message: PHP Warning:  mysqli::query(): MySQL server has gone away in /storage/www/xxxxxxx/xxx.Core/database/xxx_MySQL.class.php on line 185#012PHP message: PHP Warning:  mysqli::query(): Error reading result set's header in /storage/www/xxxxxxxx/xxx.Core/database/xxx_MySQL.class.php on line 185#012PHP message: PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 94565 bytes) in /storage/www/xxxxxxx/xxx.Core/application/xxx_Application.class.php on line 426" while reading response header from upstream, client: xxx.26.9.84, server: www.xxxxx.com, request: "GET /xxxxxx_/xxxx_.html HTTP/1.1", upstream: "fastcgi://unix:/tmp/php5-fpm.sock:", host: "failover-ip41.xxxx.de", referrer: "http://failover-ip41.xxxx.de/xxxxx/xxxxx.html"

If i switch back to older release on the other pfsense node, all seems fine again. Weird!

PiBa-NL:
I think the server and client timeouts are pretty low for mysql connections. And if your testing with low traffic, that might be part of the reason why the connection is closed early.?.

Just tested with a ssh connection and 30 second timeouts, viewing 'top' output stays alive for 12 minutes, stopping 'top' it closes the connection after the timeout happens.. So in that regard tcp connection seems to work properly.

I see your also using Transparent-ClientIP option, is it possible for you to test without that?

justnx:
The timeouts are fine for my needs. Low traffic in this case was meant, i'm clicking between Links on a not public known dev Website. On every 10 clicks i get that error exception.
The previous release was used to handle my setup with 40.000 Web concurrent connection/min with same config and without a single timeout.

You should test it with these settings i posted in a mysql environment.

PiBa-NL:
Ok ill try and install a mysql server and do some testing.
b.t.w. "Allowed memory size of 268435456 bytes exhausted (tried to allocate 94565 bytes)" seems unrelated to haproxy. The "MySQL server has gone away" could be..

justnx:
The weird is, its related to the other error, so The exhausted just happen in combination with the gone mysql connections

PiBa-NL:
I cannot find why it is causing issues for you. With your config and SQLworkbench i can run queries agains a MySql server behind haproxy as much as i want without getting any errors.. Even when returning 800000 records from a table taking 70 seconds on my virtualbox pfsense&client.. Both client and server run on windows in my case, and i dont have a php application that does the sql connections. Is that indeed the missing part to reproduce the issue?

justnx:
Have you tried that in transparent mode?

PiBa-NL:
Yes i used your config which contains 'source ipv4@ usesrc clientip' i copied that config, changed only the ip's and ports and started haproxy with it to perform my test.
b.t.w. do make sure that client and server are on different subnets and use this haproxy box your testing with as their default route as otherwise the server will not return traffic to haproxy but directly to the client which doesnt expect that to happen..

PiBa-NL:
p.s. besides starting haproxy manually with your configuration i did previously configure haproxy package on that port so that the appropriate fwd rule was shown in 'ipfw -x 4000 show'.. Might help if your testing outside the package..

PR #1118, opened 2015-10-28
snort - XMLRPC sync fixes

doktornotor:
- Fix CARP protocol/port selection
- Properly disable 'Start Snort' for CARP/HA targets
- Do some validations (IP/hostname, port, password) before attempting to sync
- Handle IPv6 addresses for sync target
- The enable checkbox was not being used at all
- The foreach loop skipped all remaining targets if one was misconfigured, due to the 'return' in there
- Do some code style cleanups and polish log messages a bit while here

### DIFF
--- a/config/snort/snort.inc
+++ b/config/snort/snort.inc
@@ -3802,11 +3802,7 @@ function snort_sync_on_changes() {
 					$protocol = $sh['varsyncprotocol'];
 					$error = '';
 					$success = TRUE;
-					if ($sh['varsyncusername']) {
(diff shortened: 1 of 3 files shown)


rbgarga:
@bmeeks8 can you check these changes please?

doktornotor:
@rbgarga - Already been dealth with via forums (for both Snort and Suricata) since this @ stuff did not exactly work last time. :)

bmeeks8:
I am OK with these fixes. Thanks @doktornotor for submitting them.

PR #1119, opened 2015-10-28
suricata - XMLRPC sync fixes round two

doktornotor:
- the CARP (auto) option has never worked due to outdated system settings location being checked
- the enable checkbox was not being used at all
- the foreach loop skipped all remaining targets if one was misconfigured, due to the 'return' in there
- polish log messages a bit


rbgarga:
@bmeeks8 Can you please check this changes before you start working on 2.0.9?

bmeeks8:
I am good with these changes. Merge them and I will then update my fork before starting on the 2.0.9 changes.

PR #1135, opened 2015-11-03
haproxy-devel, update package to use 1.6.2 binaries

PiBa-NL:


rbgarga:
I've built 1.6.2_1 PBIs. This version already have CPU_AFFINITY by default. Can you try it and update PBI version on pkg_config.10.xml?

PR #1183, opened 2015-11-18
Squid 3 support for additional ports

TheNetStriker:
This mod adds a new config page to configure additional ports for
squid. This way it is possible to e.g. add a port that requires client
certificates.


pfSense-git-updates:

doktornotor:
Why are you adding an entire duplicated page?! To add a port for Squid to listen on? What is this actually doing that cannot be done using the existing GUI by simply allowing more ports to be put into the General tab in squid_reverse_general.xml? BTW, the commit is missing the required file for 3.4.

I'd seriously appreciate if people started a discussing *beforehand* on the forums.

TheNetStriker:
I've created a thread in the forum in April, I just didn't have the time yet to commit my changes:
https://forum.pfsense.org/index.php?topic=93076.0;topicseen

I've added a new page so that multiple additional ports can be added with different ssl certificate settings. (I for instance need one ssl port without client certificates and one that forces client certificates)

Is it also possible to integrate such a "n value" page on an existing page? (I didn't find any packages that I could use as an example) Is there an example where I can see how this can be done in a page with other settings?

doktornotor:
I must be missing something here, seriously. So, you add a page that adds lots of redundant stuff -- and the only goal here to produce client certificate authentication for reverse proxy? Perhaps we could start here: https://forum.pfsense.org/index.php?topic=63138.msg459352#msg459352

The code needs to be maintainable, should the package make coffee as well?

TheNetStriker:
For one I've added the option for client certificates to better secure my web applications but I also added the possibility to configure multiple https ports with different ssl settings. The new page could theoretically replace the config section "Squid Reverse HTTPS Settings"  in the generals page, then there would be no redundant stuff. There should only be an update procedure for the https settings in the generals tab. If this would be a valid solution I will take a look at it.

TheNetStriker:
I don't know Haproxy yet. If it is possible to use that as a https reverse proxy I will take a look at it.

cbuechler:
This was summed up pretty well by @doktornotor. This is unwieldy and further overcomplicates things, for an uncommon circumstance that's better served by haproxy.

doktornotor:
I'll look into updating and getting merged the [patch linked on the forum!](https://forum.pfsense.org/index.php?topic=63138.msg459352#msg459352) that adds a very basic client cert auth fields in case someone needs such things. It's simple enough without potential break tons of other things.

PR #1193, opened 2015-11-23
Add explicit includes to a bunch of packages

doktornotor:
This is mainly in reaction to recent breakage on 2.3 caused by removing `require_once("xmlrpc.inc");` from pkg-utils.inc, but I think it's better to state the requirements explicitly rather than rely on getting the functions included indirectly from elsewhere.

(BTW, https://github.com/pfsense/pfsense/pull/2102 can be reverted once this gets merged and synced to the https://github.com/pfsense/FreeBSD-ports repo, after that none of the packages viable for 2.3 would rely on it any more.)


rbgarga:
@doktornotor could you please bump version of modified packages?

doktornotor:
@rbgarga - Done except for Squid3, I have a bunch of input validations fixes here, will do a separate PR for that and bump it there.

rbgarga:
Thank you!

PR #1194, opened 2015-11-28
Squid3 - lots of input validations, GUI logic and other fixes

doktornotor:
**Proxy - General**
- Force users to select at least one proxy or reverse proxy interface when enabling Squid (unless reverse proxy is enabled)
- Only allow to configure transparent proxy on interfaces where Squid is actually running (never had any effect otherwise anyway)
- Only allow to configure HTTPS/SSL Interception on interfaces where Squid is actually running (never had any effect otherwise anyway)
- Do not add loopback interface twice when transparent proxy is enabled and loopback is selected in Proxy Interface(s) to avoid log errors
- Avoid adding empty localnet ACL
- Fix HTTPS proxy default port
- Hide loopback from interface selection in places where it makes no sense
- Make it possible to deselect CA from SSL/MITM.
- Some code style fixes and cleanups

**Reverse Proxy**
- Add client cert auth option
- Allow reverse proxy to be used without running normal Squid proxy
- Fix no-op "Ignore Internal Certificate Validation" option
- Make sure that 'User Defined Reverse Proxy IPs' are locally configured
- Improve descriptions, cosmetics

**Antivirus**
- Switch ClamAV mirrors order to prefer manually configured mirrors

**Others**
- Fix some badly used &lt;required/&gt; XML tags all around.


marcelloc:
@doktornotor, ssl intereception can be used without transparent proxy.

doktornotor:
I sincerely hope the author of that Squid "feature" will burn in hell one day.

PR #1197, opened 2015-12-02
bind - add privileges configuration; shellcmd - nuke useless GUI tab

doktornotor:
Eh, did not intend to commit the shellcmd stuff here, but it does not matter really.


cbuechler:
Thanks @doktornotor, merged these and the others.

PR #1206, opened 2015-12-04
pfblockerng - fix check so that Alexa list is not downloaded when disabled

doktornotor:
@BBcan177 - there's something obviously fishy with the entire pfb_global() thing, you might want to re-check the whole thing.


BBcan177:
Thanks Dok for catching that... I will check over the pfb_global() function calls to see if any other variable(s) are missing. 

I will be submitting another PR soon, so better to merge this PR after that is submitted, so that we only bump the package version once.

doktornotor:
There were a couple of users complaining about it on the forum. Kinda a serious issue on boxes like Alix, the damned file eats almost half of the ramdisk. I think the whole thing should be moved out of /var altogether.

PR #1208, opened 2015-12-04
Stop syncing nmap package with FreeBSD-ports repo

doktornotor:
All PBI crap removed in https://github.com/pfsense/FreeBSD-ports/pull/4


jim-p:
FYI- I think @rbgarga said the sync script will not be run any more so this shouldn't be needed from here on. If it's needed for a single port we can run it individually rather then en masse.

doktornotor:
OK, if you are done with syncing, just close the PRs, thanks. :+1:

PR #1210, opened 2015-12-05
Squid3 - Remove the broken patch CP feature (Bug #5594)

doktornotor:
This "feature" has been broken for almost a year without anyone noticing; see https://redmine.pfsense.org/issues/5594 for details. Remove this junk until a saner alternative is available. System Patches is the way to go until that happens if someone really requires that.

On upgrade, an attempt to restore a backup will be made, and appropriate notice filed (and logged). There are also GUI instructions left in place of the original checkbox.

### DIFF
--- a/config/squid3/34/squid.inc
+++ b/config/squid3/34/squid.inc
@@ -617,7 +617,7 @@ function squid_deinstall_command() {
 /* Migrate configuration from god knows which Squid package versions */
 /* None of these ever existed with Squid 3.4 package and this cruft should be most likely just removed */
 function squid_upgrade_config() {
-	global $config;
+	global $config, $pfs_version;
 	/* migrate existing csv config fields */
 	if (is_array($config['installedpackages']['squidauth']['config'])) {
 		$settingsauth = $config['installedpackages']['squidauth']['config'][0];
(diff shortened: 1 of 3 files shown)


marcelloc:
IIRC, the patch on squid3 was fixed to 2.2 before the first "remove try". I'm not sure when it get broken again but I'll try to find it on previous squid code improvements.

The best place for it is on pfsense captive portal code to allow access to local deamon only after cp authentication.

doktornotor:
Well the thing got broken with https://github.com/pfsense/pfsense/commit/98bf4991dc31f97fc7315a6b8aba433de9d39cea which is dated Jan 15, 2015. There was nothing changed in squid regarding this. Simply noone noticed for ~11 months.

rbgarga:
It won't be automatically replicated to FreeBSD-ports anymore, so the same fix should be applied there...

jim-p:
@rbgarga It was already done there as a part of https://github.com/pfsense/FreeBSD-ports/pull/18

rbgarga:
cool, thanks!

marcelloc:
For those who needs to deny access to squid package before captive portal authentication, here is a patch to apply on pfsense 2.2.5

################################################################
--- captiveportal.225.inc       2015-12-09 18:43:50.000000000 -0200
+++ captiveportal.inc  2015-12-09 19:28:27.000000000 -0200
@@ -539,6 +539,10 @@
 EOD;

        $rulenum = 65310;
+        /* Deny direct access to local services before captive portal authentication */
+        $local_service_ports="3128,3129";
+        $cprules .= "add {$rulenum} skipto 65314 ip from any to table(100) {$local_service_ports} in\n";
+        $cprules .= "add {$rulenum} skipto 65314 ip from table(100) {$local_service_ports} to any out\n";
        /* These tables contain host ips */
        $cprules .= "add {$rulenum} pass ip from any to table(100) in\n";
        $rulenum++;
################################################################
captive_portal_patch

doktornotor:
@marcelloc - thanks, posted to the [forum topic](https://forum.pfsense.org/index.php?topic=100167.msg558141#msg558141) since the formatting here really sucks...

PR #1216, opened 2015-12-13
Add squid3 reverse-proxy security options

geijt:
Added squid3 reverse-proxy security options to control the reverse-proxy protocols and ciphers.

### DIFF
--- a/config/squid3/34/gen_dhparams.php
+++ b/config/squid3/34/gen_dhparams.php
@@ -50,15 +50,11 @@ run($argc, $argv);
 function run($argc, $argv) {
 	require_once("config.inc");
 	require_once("pfsense-utils.inc");
-	
-//	echo "input: " . $argv[2] . "\r\n";
-	
+		
 	$dhfile = $argv[2];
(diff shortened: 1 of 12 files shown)


doktornotor:
Bunch of notes here: 
- 100+ lines of code to generate DH params? Seriously? When it's already there and if someone really desires [they can generate their own in 3 lines](https://doc.pfsense.org/index.php/Importing_OpenVPN_DH_Parameters)?
- Why is disable_sslv2, disable_sslv3 even an option, plus disabled by default?
- Wouldn't a per-server (default off) checkbox with something like "Allow shitty crypto" make much more sense for cases where there's really no way around it, plus everything else HTTPS defaulting to some sensible defaults [as already done for the HTTPS proxy](https://github.com/pfsense/pfsense-packages/blob/master/config/squid3/34/squid.inc#L1164)?

cbuechler:
The DH parameters generation should go. Just use those that are there already. 

There isn't any valid reason at this point to allow use of SSLv2 and v3 at all, much less enabled by default. That should just be off with no option to enable.

PR #1224, opened 2015-12-22
Add new package surftool

reimht:


topper727:
What does it do.  Want to see it working

reimht:
Surftool manage your squidguard groups/acls. This tool was made for schools. You can set every group very easily to one of five modes. The modes on/off/only/on plus and adminfree are available. 

This package works in our school since about 1 year without any issues.

It only writes to the squidGuard configuration (mostly /usr/pbi/squidguard-amd64/etc/squidGuard/squidGuard.conf) and does not change any pfsense settings.

You can see a short description at: http://tinyurl.com/surfwiki. There is a custom repository with the base url http://pfsense.hreimers.com.

cbuechler:
2.2.x and earlier packages are no longer actively maintained, and this would need to be updated for 2.3. If you'd like to update and re-submit to the FreeBSD-ports repo for 2.3, should be able to get this merged.

PR #1225, opened 2015-12-23
haproxy-devel, small fixes for showing/hiding acl fields, fix custom action, pkg v0.36

PiBa-NL:
haproxy-devel, small fixes for showing/hiding acl fields not applicable, fix custom action, pkg v0.36

### DIFF
--- a/config/haproxy-devel/haproxy.xml
+++ b/config/haproxy-devel/haproxy.xml
@@ -42,7 +42,7 @@
 	]]>
 	</copyright>
 	<name>haproxy</name>
-	<version>0.35</version>
+	<version>0.36</version>
 	<title>HAProxy</title>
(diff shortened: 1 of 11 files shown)


rbgarga:
Are you going to send similar changes to 2.3?

PiBa-NL:
I am working on some 2.3 changes a.t.m. actually using the form classes and bootstrap stuff..
From what i read the packages should currently still sync over to the ports branche automatically.?

Next pull request will set maximum version to 2.2.999 which should stop that. If i understood correctly.

rbgarga:
We are not running sync script anymore, if you need it in this case it' s fine to call it just for this package. I was just checking to be sure

PiBa-NL:
Ok, well then it doesn't matter for me, ill be including these changes in the 2.3 pull-requests ill send in a few days.

PR #1227, opened 2015-12-29
Allow to select OSPF Network Type in the Quagga OSPFd > Interface Set…

agpecam:
…tings

To run OSPF over OpenVPN/TUN between RouterOS (client) and pfSense (server) I have to set OpenVPN-interface-network-type to 'Point-To-Point' on both sides, having "...netmask mismatch" otherways.
Two confusing things: 1) I do not know what network type is default to Quagga OSPFd, so my default is 'not specified'. 2) Name and description of my drop-down box for English is not my native language.
Thanks.

### DIFF
--- a/config/quagga_ospfd/quagga_ospfd.inc
+++ b/config/quagga_ospfd/quagga_ospfd.inc
@@ -109,6 +109,9 @@ function quagga_ospfd_install_conf() {
 			foreach ($config['installedpackages']['quaggaospfdinterfaces']['config'] as $conf) {
 				$realif = get_real_interface($conf['interface']);
 				$conffile .= "interface {$realif}\n" ;
+				if (!empty($conf['networktype'])) {
+					$conffile .= "  ip ospf network {$conf['networktype']}\n";
+				}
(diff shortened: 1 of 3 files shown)


jim-p:
In looks OK but you'll need to bump the package version in the manifest and in the package's .xml file.

Can you also submit this against the copy of the package in our freebsd-ports repository? The packages have diverged a bit so changes have to be done both places (or, preferably, only on the copy in our freebsd-ports repo)

agpecam:
Sorry, the package's .xml files have different versions in pfsense/pfsense-packages repository.
quagga_ospfd.xml - 0.6.9
quagga_ospfd_interfaces.xml and quagga_ospfd_raw.xml - 0.6.6
Should I bump them all to 0.6.10? I am in doubt.

jim-p:
We're working on simplifying all of that. For the moment only worry about bumping the Makefile.

jim-p:
(That's for the 2.3 version) -- for the pfsense-packages repo just update the main one, quagga_ospfd.xml for now

PR #1228, opened 2016-01-03
Redmine #5727 TFTP protect against empty stuff

phil-davis:
1) If the user presses Save when there was no interface selected, and
the user has still not selected an interface, then do not attempt to
unset() tftpdinterface in the config when it is already unset.
2) If the user does backup/restore or other similar stuff when the
tftpboot directory has no files in it, then do not do chmod of the
(non-existent) files. This avoids a "No such file or directory" message
being logged.
This is for 2.1.* and 2.2.* - I will also make an equivalent PR in the other repo for 2.3


phil-davis:
I don't see where the TFTP package has been converted yet for 2.3 - so there is no code to patch for that.

PR #1229, opened 2016-01-14
Logic simplifications

stilez:
1) Nested IF statements can be combined
2) Add/remove shellcmd contains logic that only needs to run if the early shellcmd array does/doesn't exist
3) Logs should be non-vague ("...added/removed a shellcmd" is better if it states the command added/removed, and for removal, clarify it removed an existing command)

### DIFF
--- a/config/systempatches/patches.inc
+++ b/config/systempatches/patches.inc
@@ -183,16 +183,11 @@ function bootup_apply_patches() {
 	$a_patches = &$config['installedpackages']['patches']['item'];
 
 	foreach ($a_patches as $patch) {
-		/* Skip the patch if it should not be automatically applied. */
-		if (!isset($patch['autoapply'])) {
-			continue;
-		}
-		/* If the patch can be reverted it is already applied, so skip it. */
-		if (!patch_test_revert($patch)) {
-			/* Only attempt to apply if it can be applied. */
-			if (patch_test_apply($patch)) {
-				patch_apply($patch);
-			}
+		/* Skip if it should not be automatically applied;
+		   only attempt to apply if it can be applied;
+		   and	if it can be reverted it is presumably already applied, so skip it. */
+		if (isset($patch['autoapply']) && patch_test_apply($patch) && !patch_test_revert($patch)) {  
+			patch_apply($patch);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


config/systempatches/patches.inc
```diff
@@ -200,36 +195,38 @@ function bootup_apply_patches() {
...
-		if (stristr($cmd, "apply_patches.php")) {
-			$found = true;
+	if (is_array($a_earlyshellcmd)) {
+		foreach ($a_earlyshellcmd as $idx => $cmd) {
+			if (stristr($cmd, "apply_patches.php")) {
+				$found = true;
+			}
 		}
+	} else {
+		$a_earlyshellcmd = array();
```
jim-p: The way the code was changed, this line (and thus the whole "else") is no longer necessary

config/systempatches/patches.inc
```diff
@@ -200,36 +195,38 @@ function bootup_apply_patches() {
...
+			}
 		}
-	}
-	if ($removed) {
-		write_config("System Patches package removed a shellcmd");
+		if ($removed) {
+			write_config("System Patches package removed existing early shellcmd: apply patches");
+		}
+	} else {
+		$a_earlyshellcmd = array();
```
jim-p: Same here. It was only making an array to skip the is_array() test earlier.

config/systempatches/patches.inc
stilez: Surely not. The one below, yes but this one adds an item to earlyshellcmd if nothing is found, and will therefore always need the array created if it didn't exist, since that's the most obvious way for !$found to be true

jim-p: It does not need to be explicitly created as an array in that case. It will be turned into an array during the assignment with []. The only reason it had to be explicitly declared was to make foreach happy.

stilez: Now that's something new to me. Did not realise that $x[] = ... would implicitly create/recreate $x as an array if it wasn't one or didn't exist

jim-p:
Somehow I overlooked that this was done against the old pfsense-packages repo, can you resubmit this against our freebsd-ports repo for the 2.3 copy of the package instead? I'd rather not make any more changes to the 2.2.x version of the package unless they're necessary.

stilez:
Against the master or devel branch?

rbgarga:
@stilez against devel please. master branch is a read-only copy of FreeBSD's repo

stilez:
Resubmitted, see https://github.com/pfsense/FreeBSD-ports/pull/47

PR #1232, opened 2016-01-24
Fix the zone generator for LOC records

cuteredstorm:
Currently all but TXT, SPF and AAAA are appended a "." when the value contains /[a-zA-Z]/ however at least the LOC record must be created without the "."


cbuechler:
This package has been revamped in the FreeBSD-ports repo for 2.3. If it's still an issue there, please re-submit the PR to that repo for the 2.3 package and we'll get it merged promptly. Thanks!

PR #1234, opened 2016-01-24
Enable master type zone when syncing to backup server

cuteredstorm:
Bind package forced "slave" type on every zone that was synced to backup server. However many servers run side-by-side and should keep the master type. I had changed the code so that when there is no master zone ip configured, the sync leaves the zone type unchanged.


cbuechler:
This package has been revamped in the FreeBSD-ports repo for 2.3. If it's still an issue there, please re-submit the PR to that repo for the 2.3 package and we'll get it merged promptly. Thanks!

PR #1236, opened 2016-02-01
haproxy, fix 'host contains' should use hdr_sub instead of hdr_dir in config

PiBa-NL:
haproxy, fix 'host contains' should use hdr_sub instead of hdr_dir in config


rbgarga:

PR #1238, opened 2016-02-10
Excessive NTLM or Negotiate auth helper annotations

gianperini:
We have switched from squid 3.2.x to 3.4.2. in our environment we are using squid with the ntlm_auth helper to do NTLM user auth against windows DC.
After switching to squid 3.4.2 squid uses nearly 100% cpu after a few minutes, with squid 3.2.x everything worked well. 
This issue was patched in the comment 28-29, in the following link -> http://bugs.squid-cache.org/show_bug.cgi?id=3997#c43.
We would like to know if its possible to apply this patch estou the current project. We don't have any experience in this kind of project, we are starting now.


pfSense-git-updates:

gianperini:
I made the subscription, what should I do next?

cbuechler:
Not sure what you're attempting there. Merging master into RELENG_2_1 isn't useful, and I'm not sure what among the 1500+ changed files and 250+ commits you're proposing. The linked bug seems to be fixed in squid 3.5, so the package ought to just be bumped to 3.5.

PR #1239, opened 2016-02-17
SquidGuard - fix ldap based groups fetch with batch mode

marcelloc:
Current version 1.4 for 2.2 is compatible with squid3 but ldap options get erros on config if enabled.
To fix it I've included a batch mode ldap fetch that update lists every 5 minutes using php/system ldap functions. 

batch_mode_ldap_options

Also "clean" the client(source) field and support multiple groups on same acl as well.
group_acls


rbgarga:

PR #1240, opened 2016-02-17
BUGFIX: named / bind : avoid journal errors with dynamic dns

WilliamHiles:
This fixes a bug with named when dynamic dns is enabled and zones are not started due to journal errors. Existing code does a "killall" which does not allow named to flush the zone and journals to disk. This results in zone files files that are out of sync causing named to fail loading the zone and for name resolution to fail.  A bad thing.

Fix: rc_stop to use the isc recommended method for shutdown.
Fix: rc_start no longer allows multiple instances to be accidentally started
Add: rc_start detects and clears the journal when journal errors occur during startup.
These errors are critical for users running dynamic dns.

Recommend adding this to 2.2.

### DIFF
--- a/config/bind/bind.inc
+++ b/config/bind/bind.inc
@@ -312,6 +312,7 @@ EOD;
 		$bindview = array();
 	}
 
+	$master_ddns_zones = "";
 	for ($i = 0; $i < sizeof($bindview); $i++) {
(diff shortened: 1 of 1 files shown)


pfSense-git-updates:

WilliamHiles:
CLA was submitted a while ago.  No idea why it's not showing up.

PR #1242, opened 2016-03-15
Check for libldap shared objects needed by squidGuard

cedriczirtacic:
As it says in the squidGuards docs: "In order to use LDAP functionalities the system must have the proper LDAP libraries and include files installed (openldap works fine)."


cbuechler:
openldap-client is a dependency of the squid package, so it's always there. If you have a situation on 2.3 where squid and squidguard are installed and you don't have openldap-client, please post details to the forum. 

This is also the 2.2.x and earlier package repo, which don't have is_pkg_installed as a function, so this would just break things.

PR #1245, opened 2016-03-22
Bacula Client:Allow remote director with local named filedaemon

Tommi2Day:
Its very unlikely a local director is available on a pfsense box. Usually you will have a remote director and a filedaemon which may get the hostname in its name. This patch will offer such configuration option


pfSense-git-updates:

rbgarga:

Tommi2Day:
sure, if you wait a half year to check it

PR #1248, opened 2016-04-05
Fixing evaluation of some pool configuration items

darksoul42:
With the previous evaluation mechanism, setting retries to 0 would get evaluated as uninitialized, so set to default 3.
This is a huge problem in cases where HAProxy should NOT retry at all.

Given that items in $pool should already be string-ified, evaluation boils down to whether it's an empty string or not, without consideration of nullness.


pfSense-git-updates:

cbuechler:
Thanks @darksoul42. This has been merged to the 2.3+ package branches.

PR #1252, opened 2016-04-25
Fix certs from external CAs not being usable to create clients

ikogan:
The certificates I use, generated by FreeIPA and imported into pfsense, are currently not usable when exporting clients as, internally, there is no caref from an imported certificate to an imported CA.

This changes the behavior of the client export package a little to find certificates where their issuer matches the subject of the VPN server's CA certificate.


pfSense-git-updates:

ikogan:
CLA has been signed, thanks.

rbgarga:

PR #1253, opened 2016-05-04
Fix against code execution attack

cedriczirtacic:
There isn't a check against an improper of user input (blacklist URL) and PHP code (thus system commands) could be executed as root. Example of a _special_ URL: `","");passthru('/usr/bin/id > /tmp/id');//`

### DIFF
--- a/config/squidGuard/squidguard_blacklist.php
+++ b/config/squidGuard/squidguard_blacklist.php
@@ -64,14 +64,12 @@ function squidguard_blacklist_AJAX_response( $request )
 
     # Actions
     if ($request['blacklist_download_start'])
-        {
-                $url = $request['blacklist_url'];
-                if ( preg_match('/["\'\s\n\r\(\);]+/', $url )){
-                        $status = "Error: Bad blacklist URL";
-                        squidguard_update_log($status,"");
-                }else{
-                	squidguard_blacklist_update_start( $url ); # update start
-                }
+    {
+    	if( squidguard_blacklist_update_start( $request['blacklist_url'] ) === false ) # update start
+    	{
+    		$status = "Blacklist update error. Please check URL";                                                                                                                  
+		squidguard_update_log($status,"");
+    	}                                                                                                                                                                          
     }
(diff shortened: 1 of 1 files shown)


config/squidGuard/squidguard_blacklist.php
```diff
@@ -58,11 +58,19 @@
...
+    $status = '';
     $sz  = 0;
     $pcaption = '&nbsp;';
 
     # Actions
-    if     ($request['blacklist_download_start'])  squidguard_blacklist_update_start( $request['blacklist_url'] ); # update start
+    if ($request['blacklist_download_start'])
+        {
+                $url = $request['blacklist_url'];
+                if ( preg_match('/["\'\s\n\r\(\);]+/', $url )){
```
jim-p: A better test is to use:
`if (filter_var($url, FILTER_VALIDATE_URL) === false)`

config/squidGuard/squidguard_blacklist.php
```diff
@@ -58,11 +58,19 @@
...
+                        $status = "Error: Bad blacklist URL";
+                        squidguard_update_log($status,"");
+                }else   squidguard_blacklist_update_start( $url ); # update start
```
jim-p: See the style guide for recommendations here -- always use curly braces in these cases. The code in that file is a bit of a mess but it's still better to make things right as it's edited.

jim-p:
Looks like a good fix but it should probably be fixed at the root of the problem:
1. Validate on save in the GUI and don't save an invalid URL
2. Validate the URL at the start of squidguard_script_blacklistupdate() before it gets put into the script, that way it can be protected no matter who calls the function.

cedriczirtacic:
Thanks for the positive comments, jim-p!
Some things that you pointed out:

1. Also think that it should be validated in the GUI via JS but I don't think it gets saved because _squidguard_blacklist_update_start()_ isn't called if the regex check failed.
2. Yes, that is a better idea. I will take a look tomorrow and propose a change on that function.

rbgarga:

PR #1254, opened 2016-05-05
Blacklist URL validation

cedriczirtacic:
URL validation added to added to _squidguard_blacklist_update_start()_. (View pull #1253)


rbgarga:

PR #1256, opened 2016-05-30
Add wildcard on vhosts package if host empty

cdesbouche-acipia:
Add wildcard on vhosts package if host empty.

Currently, this package always try to define a equal ("==") condition.
If host is defined empty, config file for this host will switch to regex condition "=~" with a wildcard regex "^(.*)$"


pfSense-git-updates:

rbgarga:

PR #1262, opened 2016-08-24
Update ntopng.inc to add ipv6 "local network" addresses

garyd9:
For each of "selected", "lanonly" and "rfc1918" options on describing local networks, add the appropriate IPv6 addresses (if any) in addition to the IPv4 address.  If this is applied to an existing ntopng installation, the historical data will have to be reset in order to the existing "local" ipv6 addresses to be recognized as local.


garyd9:
First pull for pfsense...  (and just completed ICLA)

rbgarga:

PR #1263, opened 2016-09-16
Strip carriage returns from 'raw' openbgpd config

Xon:
On Windows, copy & pasting text into the raw config text box can include carriage returns. This causes OpenBGPd to error unexpectedly with a generic "syntax error"


Xon:
I thought this worked at some stage, but I'm unsure where the best places is to scrub out carriage returns.

pfSense-git-updates:

rbgarga:

PR #1266, opened 2017-02-04
fix wrong description

FlorianHeigl:
it says "blocked domains" but for all I can tell this was copypasted from somewhere.


netgate-git-updates:

FlorianHeigl:
i do not < /care >

PR #1268, opened 2017-08-02
Table BGColor

NonSecwitter:
the table is unreadable with current color settings and pfSense portal in dark theme.

Here are the colors I've tested: https://snag.gy/VFxeiK.jpg
#919191 is the line with "[blk_BL_automobile_bikes]"
Note that the word "access" at the end of the line is black to reflect light theme colors.

The lowest row, just above the developer tools, is the original presentation. The text is unreadable.

### DIFF
--- a/config/squidGuard/squidguard.inc
+++ b/config/squidGuard/squidguard.inc
@@ -678,7 +678,7 @@ function make_grid_general_items($id = '')
 {
     global $squidguard_config;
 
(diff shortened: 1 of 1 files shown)


rbgarga:
This repository is obsolete. Packages are handled in https://github.com/pfSense/FreeBSD-ports repo since 2.3

PR #1270, opened 2018-01-21
Updating some verbiage

vrtadmin:
Updating some verbiage here.  Changing "Snort VRT" to "Snort Subscriber Rule Set" and "VRT" to "Talos" where appropriate.  Also added in the "Max-Detect" policy present in the ruleset.


rbgarga:
This repository contain legacy coded used by pfSense < 2.3. If you want to submit changes for current packages use pfsense/FreeBSD-ports

